PR #5, opened 2022-07-25
In convertMedia.ts, images and illustrations are optional

chrisvire:
* Before, if there were no images, then the conversion would fail
  with a trackback in the output.
* Add illustrations as optional media files to process


chrisvire:
Already added by beb5d1c9e42c318a65774542681f603225d47c04

PR #36, opened 2022-09-08
Update convertAbout.ts

chrisvire:
* SAB has been changed to generate about.partial.html in the data directory. It should be copied to the static folder.


chrisvire:
From @7dev7urandom in Discord:

> Looks good to me, if you just want the file dropped into static as-is

7dev7urandom (approved):

PR #48, opened 2022-09-26
Fix clean

chrisvire:


chrisvire:
@7dev7urandom Is there a reason you didn't include `static/collections` ?

chrisvire:
It is in this pull request.

PR #49, opened 2022-09-29
Enable Chapter Selector based on features

chrisvire:
* mainFeatures: show-chapter-number-on-app-bar - controls whether the number displays on the app bar
* mainFeatures: show-chapter-selector - controls whether the chapter can be changed by a selector if the number displays on the app bar

Note: currently the navigation is not working due to depending on the chapter selector being present.

Todo: Change the navigations so that it uses a route (and the address of the page changes) instead of depending on the controls.


chrisvire:
This will be done in #80

PR #52, opened 2022-10-02
Fix Convert Config

chrisvire:
* Add verbosity level to help with tracking down issues
* Fix writing-system not required to have display-names
* Fix font-name not required for font
* Fix len and size not required for audio filename

### DIFF
--- a/scripts/convertConfig.ts
+++ b/scripts/convertConfig.ts
@@ -149,7 +149,7 @@ function parseConfigValue(value: any) {
     return value;
 }
 
-function convertConfig(dataDir: string, verbose: boolean) {
+function convertConfig(dataDir: string, verbose: number) {
     const dom = new jsdom.JSDOM(readFileSync(path.join(dataDir, 'appdef.xml')).toString());
     const { document } = dom.window;
 
@@ -244,9 +244,11 @@ function convertConfig(dataDir: string, verbose: boolean) {
     data.bookCollections = [];
 
     for (const tag of booksTags) {
+        if (verbose > 1) console.log(`Converting Collection: ${tag.id}`);
         const featuresTags = tag.querySelector('features[type=bc]')?.getElementsByTagName('e');
         if (!featuresTags) throw 'Book collection feature tags missing';
         const features: any = {};
+        if (verbose > 1) console.log(`. features`);
         for (const feature of featuresTags) {
             features[feature.attributes.getNamedItem('name')!.value] = parseConfigValue(
                 feature.attributes.getNamedItem('value')!.value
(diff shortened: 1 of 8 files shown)


chrisvire:
@7dev7urandom Does this look ok? Do you have any suggestions for changes?

7dev7urandom (commented):
Not a big fan of verbosity being a number, but if you're going to do that I'd either use a namespace/class to handle logging at different levels. Then verbocity doesn't have to be passed to the run function and it can just call Logger.log or Logger.dump or whatever instead of doing an if statement each time. 

Not a big deal though, looks good to me otherwise

scripts/convertConfig.ts
```diff
@@ -302,9 +313,12 @@ function convertConfig(dataDir: string, verbose: boolean) {
...
             languageCode: writingSystem.attributes.getNamedItem('code')!.value,
-            languageName: writingSystem
-                .getElementsByTagName('display-names')[0]
-                .getElementsByTagName('form')[0].innerHTML,
+            languageName:
+                writingSystem.childElementCount > 0
+                    ? writingSystem
+                          .getElementsByTagName('display-names')[0]
+                          .getElementsByTagName('form')[0].innerHTML
+                    : '',
```
7dev7urandom, line 321:
```suggestion
            languageName: writingSystem
                .getElementsByTagName('display-names')[0]
                ?.getElementsByTagName('form')[0].innerHTML,
```
In JS, accessing an element of an array that does not exist returns `undefined` instead of throwing an error as it does in Java.
Using `?` is sufficient to use [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) to set `languageName` to `undefined` if the `display-names` do not exist (making the array length 0 and `array[0]` undefined).

7dev7urandom (dismissed):

scripts/convertConfig.ts
chrisvire: Thanks. I will try that out.

chrisvire: It turns out that we require languageCode to be defined for Proskomma, but languageName is not. I did use this syntax on languageName.

scripts/convertConfig.ts
```diff
@@ -43,7 +43,7 @@ type BookCollection = {
         verseNumbers: string;
     };
     languageCode: string;
-    languageName: string;
+    languageName?: string;
```
chrisvire, line 46: @7dev7urandom Is this the right way to specify that languageName is optional?

scripts/convertConfig.ts
```diff
@@ -62,7 +62,7 @@ export type ConfigData = {
     name?: string;
     mainFeatures?: any;
     fonts?: {
-        name: string;
+        name?: string;
```
chrisvire, line 65: @7dev7urandom Should I set this as optional to document that it might not be there?

scripts/convertConfig.ts
```diff
@@ -268,8 +268,8 @@ function convertConfig(dataDir: string, verbose: number) {
                 audio.push({
                     num: parseInt(page.attributes.getNamedItem('num')!.value),
                     filename: fTag.innerHTML,
-                    len: parseInt(fTag.attributes.getNamedItem('len')!.value),
-                    size: parseInt(fTag.attributes.getNamedItem('size')!.value),
+                    len: parseInt(fTag.attributes.getNamedItem('len')?.value || ''),
```
chrisvire, line 271: @7dev7urandom Is this an OK way to handle `len` and `size` not being set?

scripts/convertConfig.ts
```diff
@@ -14,8 +14,8 @@ type HTML = string;
 type BookCollectionAudio = {
     num: number;
     src: string;
-    len: number;
-    size: number;
+    len?: number;
```
chrisvire, line 17: If I have these set as optional, then the values in the JSON are `null` (to me indicates missing) as apposed to `0` (to me indicates zero length).

7dev7urandom: Undefined not null but yes

7dev7urandom (approved):
Looks good to me

PR #54, opened 2022-10-06
Drawer Content Customization

jakedcolbert:
Search, History, Bookmarks, Notes, and Highlights will display in drawer only if they are enabled in scripture app builder.


jakedcolbert:
Share App button turned invisible if none of the sharing options are selected in scripture app builder? How will the button handle App Store redirection or Google Play redirection? What should I implement for 'Share App'

src/lib/components/Sidebar.svelte
```diff
@@ -22,6 +22,12 @@ The sidebar/drawer.
         document.activeElement.blur();
     };
     beforeNavigate(closeDrawer);
+
+    let showSearch = $globalConfig.mainFeatures['search'];
+    let showHistory = $globalConfig.mainFeatures['history'];
+    let showBookmarks = $globalConfig.mainFeatures['annotation-bookmarks'];
+    let showNotes = $globalConfig.mainFeatures['annotation-notes'];
+    let showHighlights = $globalConfig.mainFeatures['annotation-highlights'];
```
jakedcolbert, line 30: Are annotation-bookmarks, annotation-notes, and annotation-highlights exclusively settings for the annotation menu or should they be modifying the drawer too?

chrisvire: @jakedcolbert You can build an Android app to check.

chrisvire:
> * Should the Share App button be turned invisible if none of the sharing options are selected in scripture app builder?

If there are no share options enabled, then the share menu is not visible.

> * How will the button handle App Store redirection and Google Play redirection if both are toggled?

If there are multiple, show a popup to select which one.  You can see how Android does this.  Actually, Android doesn't show a link for Apple Store.  For PWA, I would like you to show options for all. Look for images on the web (try to use svg and see icons in `src/lib/icons`).

Apple App Store Guidelines: https://developer.apple.com/app-store/marketing/guidelines/
Google Play Guidelines https://play.google.com/intl/en_us/badges/

These are formatted URLs.

Google Play:  https://play.google.com/store/apps/details?id=*package* (found in package root element)
Apple Store: https://apps.apple.com/us/app/id*share-apple-id* (found in features)

>   
>   * Is that handled in the sidebar component code?

What I would suggest is that there is a service for copying data to the clipboard.  The sidebar can assign the url to a store with "displayname" and "text" properties.  A service would be listening for updates to the store and would put the data into the clipboard and create a toast (from DaisyUI) notifying the user that "displayname set in clipboard"

src/lib/components/Sidebar.svelte
chrisvire: They modify the drawer as well.

chrisvire:
Make sure to have prettier running when you are making changes to file (or run `npm run format` before committing).

jakedcolbert:
Understood. Thanks Chris, I'll make those changes now.

chrisvire:
I would suggest that what happens when share is pressed would go in another PR.

jakedcolbert:
Agreed, I'll start a new PR for the share screen.

src/tailwind.css
```diff
@@ -1,3 +1,3 @@
 @tailwind base;
 @tailwind components;
-@tailwind utilities;
+@tailwind utilities;
```
chrisvire, line 3: @jakedcolbert Did you mean to change src/tailwind.css? It looks like you deleted the trailing newline from the file and prettier doesn't like this.  To fix this:

1. run `npm run format`
2. check in the changes.

jakedcolbert: That was a mistake, I'll make a habit of running `npm run format` at each commit. Thanks

chrisvire:
@jakedcolbert Please verify that the dividers are the same as the Android app.

jakedcolbert:
@chrisvire As far as I can tell, divider visibility behaves same as the Android app.

chrisvire (changes requested):
If Notes, Bookmarks, and Highlights are turned off, then there should be a separator between history and share app.

jakedcolbert:
Thank you for catching that, fixed now.

chrisvire:
@jakedcolbert: you need to use prettier in VSCode or from `npm run format` before submitting.

chrisvire (approved):
Looks good.  Thanks for the rework.

PR #58, opened 2022-10-07
Make BC attributes optional and revert to id for abbr

chrisvire:
The BookCollection has attributes (collectionName, collectionAbbreviation, collectionDescription) that are actually not required to be filled out in the App Builder. 

The Proskomma Freeze process (in convertBooks.ts) depended on collectionAbbreviation being set.  It now falls back to id.

### DIFF
--- a/scripts/convertBooks.ts
+++ b/scripts/convertBooks.ts
@@ -36,7 +36,7 @@ export async function convertBooks(
             console.warn(`Language ${lang} already used in another collection. Proceeding anyway.`);
         }
         usedLangs.add(lang);
-        const abbr = collection.collectionAbbreviation;
+        const abbr = (collection.collectionAbbreviation ?? collection.id).toLowerCase();
         const docSet = lang + '_' + abbr;
         if (verbose)
             console.log('converting collection: ' + collection.id + ' to docSet: ' + docSet);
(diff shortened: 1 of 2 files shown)


scripts/convertBooks.ts
```diff
@@ -36,7 +36,7 @@ export async function convertBooks(
             console.warn(`Language ${lang} already used in another collection. Proceeding anyway.`);
         }
         usedLangs.add(lang);
-        const abbr = collection.collectionAbbreviation;
+        const abbr = (collection.collectionAbbreviation ?? collection.id).toLowerCase();
```
7dev7urandom, line 39: Proskomma has strict rules on the `docSet` value that (as far as I know) collection id might not meet. This works as long as it's been tested against Proskomma.

scripts/convertConfig.ts
```diff
@@ -296,7 +296,9 @@ function convertConfig(dataDir: string, verbose: number) {
             if (verbose >= 3) console.log(`.... book: `, JSON.stringify(books[0]));
         }
         const collectionNameTags = tag.getElementsByTagName('book-collection-name');
-        const collectionName = collectionNameTags.length > 0 ? collectionNameTags[0].innerHTML : '';
+        const collectionName = collectionNameTags[0].innerHTML.length
+            ? collectionNameTags[0].innerHTML
+            : undefined;
```
7dev7urandom, line 301: This will error if the collection name tag is completely missing instead of empty. Is it ever possible for the `<book-collection-name>` tag to be completely left out of the xml? 

Ditto on lines 318 and 323.

PR #69, opened 2022-11-04
Restructure Navbar

davidmoore1:
Change navbar so that it can be different from screen to screen
For the present, eliminate all navbar buttons except for the scripture view screen
Add titles to center of navbar
Change structure of navbar component so that when it is used the developer can specify right buttons, left buttons, and center items, any of which can be empty.

Also clean up Book Selector and Chapter Selector button texts so values always reflect the scripture being displayed


chrisvire (approved):
I can approve this so that other teams can add their stuff based on this structural change.  Please look into:
* moving hamburger icon to Navbar.svelte
* remove `for={drawerName}` 
* remove `<div>`s that are empty (or comment out)?
* look into `lang="ts"` for other components.

PR #72, opened 2022-11-07
Drawer image

jakedcolbert:


vite.config.js.timestamp-1667841673769.mjs
```diff
@@ -0,0 +1,8 @@
+// vite.config.js
```
chrisvire, line 1: What is this file?  Is it a generated temporary file? Does it need .gitignore'd?

src/lib/components/Sidebar.svelte
```diff
@@ -45,7 +45,10 @@ The sidebar/drawer.
         <label for={drawerId} class="dy-drawer-overlay" />
         <ul class="dy-menu p-1 overflow-y-auto w-3/4 sm:w-80 bg-base-100 text-base-content">
             <!-- Sidebar content here -->
-            <li><a href="/"><BibleIcon /><span>{$globalConfig.name}</span></a></li>
+            <a class="fill" href="/">
```
chrisvire, line 48: What happens if there is no nav_drawer.png or nav_drawer@2x.png specified?

chrisvire:
Comparing screenshot from Android and Chrome when in phone simulation mode.

Chrome (Pixel 5)

image

Android (Pixel 3XL)

image

It feels a little wide, but it is probably fine.

src/lib/components/Sidebar.svelte
chrisvire: This is what happens:

image

chrisvire (changes requested):
1) Please handle not having a Navigation Drawer image.  The Android App uses a default app icon.  You could use the Icon that was there before?

2) Figure out where the vite.config.js.timestamp-XXXX.mjs came from and stop it being generated (or at a minimum .gitignore files like it -- would need a file pattern).

src/lib/components/Sidebar.svelte
chrisvire: I like having an icon and the App Name if there is no Drawer Image.  Do you want me to try to get the App Icon or just use the icon we were using before?

vite.config.js.timestamp-1667841673769.mjs
jakedcolbert: looks like it came from the npm run format

src/lib/components/Sidebar.svelte
chrisvire: I talked it over with another developer and he said the back end should make sure there is a file.  So this is OK.

vite.config.js.timestamp-1667841673769.mjs
chrisvire: So remove this from the PR and it should be good to go.

src/lib/components/Sidebar.svelte
```diff
@@ -39,13 +39,18 @@ The sidebar/drawer.
     <input id={drawerId} type="checkbox" class="dy-drawer-toggle" />
     <div class="dy-drawer-content flex flex-col bg-gray-200">
         <!-- Page content here -->
-        <slot />
+        <slot d />
```
chrisvire, line 42: Why is this here?

PR #74, opened 2022-11-08
Menu items

jakedcolbert:


jakedcolbert:
Needs dynamic icons, fixed resolutions for now.

src/lib/components/Sidebar.svelte
```diff
@@ -74,6 +76,21 @@ The sidebar/drawer.
             <!-- svelte-ignore a11y-click-events-have-key-events -->
             <li><a on:click={closeDrawer}><TextAppearanceIcon />Text Appearance</a></li>
             <div class="dy-divider m-1" />
+            {#if menuItems}
+                {#each menuItems as item}
+                    <li>
+                        <a href={item.link['default']}>
+                            <img
+                                src="icons/{item.images[2].file}"
```
chrisvire, line 84: You need to update to the most recent SAB (build 10-Nov 2022). The location of the menu item icons is now `icons/menu-items/` so that user-created icons don't conflict in name with icons that might be delivered with the app.

src/lib/components/Sidebar.svelte
```diff
@@ -74,6 +76,21 @@ The sidebar/drawer.
             <!-- svelte-ignore a11y-click-events-have-key-events -->
             <li><a on:click={closeDrawer}><TextAppearanceIcon />Text Appearance</a></li>
             <div class="dy-divider m-1" />
+            {#if menuItems}
+                {#each menuItems as item}
+                    <li>
+                        <a href={item.link['default']}>
+                            <img
```
chrisvire, line 83: Did you try to use the `<picture>` element that can handle multiple resolutions (especially since the extra icons are 2x and 3x the size).

chrisvire (changes requested):

src/lib/components/Sidebar.svelte
```diff
@@ -74,6 +75,27 @@ The sidebar/drawer.
...
+                    <li>
+                        <a href={item.link['default']}>
+                            <picture>
+                                <source
+                                    srcset="icons/menu-items/{item.images[1]
+                                        .file} 2x, icons/menu-items/{item.images[2].file} 3x"
+                                />
+                                <img
+                                    src="icons/menu-items/{item.images[0].file}"
+                                    height="20"
```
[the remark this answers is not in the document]
chrisvire: Let's create a separate issue to get all of the icons at 24x24 and document like MoreVertIcon.svelte.  Include changing this icon size at the same time.

PR #81, opened 2022-11-19
Add stores to allow text size and line spacing changes to be viewed

davidmoore1:
Adding store variables that will allow the text size and line height to be changed immediately.
ScriptureRender is being rewritten for Sofia at the moment, so where these variables may be used in the final version may differ.  But the store that's affected, and will need to be updated by the text view popup, will be the same so that the team working on that feature can see the text change immediately.

To change the font size or line height from a component:
    import {
        bodyFontSize,
        bodyLineHeight
    } from '$lib/data/stores';

            $bodyFontSize = '30px';
            $bodyLineHeight = '300%';

Still to be done (not in this PR): get initial values from configuration

### DIFF
--- a/src/lib/components/ScriptureView.svelte
+++ b/src/lib/components/ScriptureView.svelte
@@ -94,10 +94,10 @@ TODO:
     $: updateHighlight($audioHighlight);
 
     let fontSize = '17px';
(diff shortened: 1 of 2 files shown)


chrisvire (approved):
Looks good to me.

PR #99, opened 2022-12-11
Added Store Function

jakedcolbert:
Added store function for converting style into a single block a styletag can use. Additionally making some edits so Collection Menu would not inherit styles from navbar.

### DIFF
--- a/src/lib/components/CollectionSelector.svelte
+++ b/src/lib/components/CollectionSelector.svelte
@@ -1,6 +1,6 @@
 <!--
 @component
-The navbar component.
+Book Collection Selector component.
 -->
 <script>
     import Dropdown from './Dropdown.svelte';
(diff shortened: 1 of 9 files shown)


jakedcolbert:
I'll need some pointers here. The collection menu makes the webpage expand horizontally. How do I format it to 'unfold' left instead of right?

src/lib/components/CollectionSelector.svelte
```diff
@@ -29,33 +29,36 @@ The navbar component.
...
+<div class="dy-dropdown dy-dropdown-end">
+    <Dropdown>
+        <svelte:fragment slot="label">
+            <BibleIcon color="white" />
+        </svelte:fragment>
+        <svelte:fragment slot="content">
+            <TabsMenu
+                options={{
+                    'Single Pane': {
+                        tab: { component: SinglePaneIcon },
```
jakedcolbert, line 44: How do I make these icons white?

jakedcolbert:
To Do:

- Update localStorage.refs to include Book Collection as docSet
- Implement:
  - Two Pane Menu 
  - Verse by Verse Menu
  - Book colorings by book type - needs to be dynamic by theme

jakedcolbert:
I'm realizing too late this PR is not clean... Hopefully it won't cause any trouble.

jakedcolbert:
Just Dropdown Styling now

jakedcolbert:
To Do: 
- Grid View elements need to be dynamically colored

PR #106, opened 2023-01-17
Audio advancement

elijahbigk77:


src/lib/components/ScriptureView.svelte
```diff
@@ -22,6 +22,7 @@ TODO:
     } from '$lib/data/stores';
     import { renderDoc } from '../scripts/render';
     import { LoadingIcon } from '../icons';
+    import Hammer from 'hammerjs';
```
chrisvire, line 25: What is Hammer for?

chrisvire: Please remove hammerjs if it is not being used.

package.json
```diff
@@ -53,6 +53,7 @@
         "eslint-plugin-prettier": "^4.0.0",
         "hammer": "^0.0.5",
         "hammerjs": "^2.0.8",
+        "jshint": "^2.13.6",
```
chrisvire, line 56: Why was jshint added?  It is related to hammer and hammerjs being added? What warning?

chrisvire:
@elijahbigk77 is this ready for review?

PR #123, opened 2023-02-16
Implement play icons

chrisvire:
* fix the audio bar layout so that different sized play icons will work
* add additional play icons
* add switching play icon based on config


davidmoore1 (approved):
Reviewed code. Looks good, ran on my Mac and it looks good there as well.

PR #126, opened 2023-02-21
Move skip function to referenceStore

chrisvire:


src/lib/data/store-types.js
```diff
@@ -147,5 +161,11 @@ export const groupStore = (/**@type{any}*/ groupType, /**@type{any}*/ props) =>
...
     };
 
-    return { subscribe, set };
+    const skip = (val, key = 'default', mod = undefined) => {
+        stores[key].skip(val);
+        mods[key] = mod;
+        subs[key].forEach((sub) => sub(vals[key], mods[key]));
+    }
+
+    return { subscribe, set, skip };
```
chrisvire, line 170: @FyreByrd I moved the `skip()` functionality to `referenceStore` (so that it can be used by swiping as well), however, I had to patch it through `groupStore` (and notify the subs).  Any thoughts on a better way to do this?

FyreByrd: I made some changes to generalize the process of exposing extra needed functionality through groupStore and opened a PR into the branch in your repo.

FyreByrd (approved):
This looks fine as is. I could not find any better solution to the issue when I tried. I did however take steps to generalize the process which I put in my own pull request into your work.

PR #132, opened 2023-03-02
Make select text reactive

davidmoore1:


chrisvire (commented):
LGTM

I have concerns with having the selectedVerses subscription in verseSelectUtil.ts. We will see how i goes.

chrisvire (approved):

PR #134, opened 2023-03-04
Delete old files before copying new files in convertMedia.ts

7dev7urandom:
Rewrote most of the `convertMedia.ts` file. Files in `static/{media folders}` are now deleted before they are copied. I also added incremental compilation, so that only modified folders will be copied.


chrisvire (approved):
LGTM. I will keep an eye on any issue with it failing.

chrisvire:
@7dev7urandom Do you want to look into working around the "FIXME" issue you documented or just go with this for now?

7dev7urandom:
I did look into it for awhile but short of adding an intentional delay I wasn't able to really fix it. When I moved the project to a different disk, I had no issues. I think it works fine, that comment is just there in case anybody ends up trying to fix the same issue. It can be merged

PR #142, opened 2023-03-08
Audio playback

elijahbigk77:


chrisvire (changes requested):
@elijahbigk77: It looks really good.  Nice job. Please make the following changes:
1. Please put this into a separate component instead of directly in the AudioBar component. If you need help with this, let me know.
2. If the sidebar is up, the left needs to be adjusted so that it is centered over the content area. Otherwise, the controls can be hidden under the sidebar
3. When in dark mode, the title should have the color #e0e0e0 (see routes/settings/+page.svelte in the main branch for how I handled it there)

chrisvire:
Here is demonstration of issue 2 above:

image

src/lib/components/AudioBar.svelte
```diff
@@ -9,6 +9,24 @@ TODO:
     import { AudioIcon } from '$lib/icons';
     import { refs, audioHighlight, audioActive, s, playMode } from '$lib/data/stores';
     import config from '$lib/data/config';
+    import { createEventDispatcher } from 'svelte';
+
+    const dispatch = createEventDispatcher();
```
chrisvire, line 14: I don't see this being used.

chrisvire (changes requested):
Good job putting the functionality into a component.

Please rename "AudioPlayback.svelte" to "AudioPlaybackSpeed.svelte" and then I will approve.

chrisvire (approved):
LGTM

PR #143, opened 2023-03-09
Add support for display verse number feature

davidmoore1:


chrisvire (approved):
Very nice. Good job with the viewSettings.  We can work on removing other store dependencies from ScriptureViewSofria

PR #152, opened 2023-03-18
Modal Component

jakedcolbert:
Adding support for modals in the PWA. A unique ID will need to be assigned to each since a label button corresponds to the physical pop up.


jakedcolbert:
Screenshot 2023-03-20 at 11 23 30 AM
@chrisvire I do not have that. But I will notify you.

chrisvire (approved):

PR #154, opened 2023-03-27
View enhancements

davidmoore1:
This became over large.  Each commit should probably have been a separate pr.
Fixed a small issue where scrolling was set for the wrong div so audio scrolling wasn't working.
Moved the remainder of the store references so that they were passed in for ScriptureViewSofria
Fixed a small issue where audio highlighting failed after going to a non audio chapter and back
Reimplemented parsePhrases to more closely model the android version
Implemented one verse per line view setting.

When reviewing, it may be easier to view each commit individually.

### DIFF
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -22,6 +22,7 @@
     import TextAppearanceSelector from '$lib/components/TextAppearanceSelector.svelte';
     import config from '$lib/data/config';
     import ScriptureViewSofria from '$lib/components/ScriptureViewSofria.svelte';
+    import { getFeatureValueString } from '$lib/scripts/configUtils';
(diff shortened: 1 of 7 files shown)


src/routes/+page.svelte
```diff
@@ -34,13 +35,19 @@
         (refs as any).skip(event.detail.direction === 'right' ? -1 : 1);
     }
 
+    $: audioPhraseEndChars = getFeatureValueString(
+        'audio-phrase-end-chars',
+        $refs.docSet.split('_')[1],
```
chrisvire, line 40: This can be `$refs.collection`

davidmoore1 (commented):
Made all the required changes and Chris approved

chrisvire (approved):
LGTM

PR #167, opened 2023-03-31
Feature/text selection toolbar

saidbrandonsaid:
Starter copy, bookmark, highlight, and note functionality


chrisvire (changes requested):
There are a lot of commits and there are conflicts.  It might be good to sqush these changes down to a single commit which will make it much easier to rebase off of main.

package.json
```diff
@@ -55,5 +55,8 @@
     },
     "volta": {
         "node": "16.17.0"
+    },
+    "dependencies": {
+        "svelte-french-toast": "^1.0.3"
```
chrisvire, line 60: move to devDependencies

chrisvire:
We merged this code in #170

PR #169, opened 2023-04-02
Book Collection selector

jakedcolbert:


chrisvire (changes requested):
These changes are needed:
1. display the description under the name if present
2. do not include collections in the single pane list if "bc-allow-single-pane" is false
3. when navigating between collections, if the previous book/chapter in not in the new collection, then go to the first book/chapter in the new collection.

jakedcolbert:
Changes completed:

- [ ] Display the description under the name if present
- [ ] Do not include collections in the single pane list if "bc-allow-single-pane" is false
- [ ] When navigating between collections, if the previous book/chapter in not in the new collection, then go to the first book/chapter in the new collection.

chrisvire (approved):
It is a good start.

PR #175, opened 2023-04-06
Fix issues in Book Selector

chrisvire:
Issues:

- [x] If there are lest than 6 cells in a group, then they are spaced out evenly instead of being in the first 3 columns
- [x] labels for books without abbreviations wasn't working (need to revert to name)
- [x] books before and after the "testaments" where not displayed in the order in the config.js.
- [x] introductions were not displaying at the top of the chapter selector.
- [x] selecting a book without chapters but had introduction was not navigating to book immediately.

Grid CSS for the win!

Before:

image

After:

image


chrisvire:
Navigation isn't working yet.

PR #178, opened 2023-04-11
Audio advances to any specific location on the progress bar when that…

elijahbigk77:
… location is clicked


chrisvire (approved):
Clean PR with only the changes needed

PR #179, opened 2023-04-12
Make changes to handle glossary and introduction books

davidmoore1:


src/lib/components/ScriptureViewSofria.svelte
```diff
@@ -222,7 +223,8 @@ TODO:
                     workspace.phraseDiv = div.cloneNode(true);
                     workspace.phraseDiv = startPhrase(workspace);
                 } else {
-                    workspace.phraseDiv = div;
+                    workspace.phraseDiv = div.cloneNode(true);
+                    console.log('last phraseDiv', workspace.phraseDiv);
```
chrisvire, line 227: Keep the console.log?

src/lib/components/ScriptureViewSofria.svelte
```diff
@@ -374,11 +381,11 @@ TODO:
...
-                            /* console.log(
-                                'Start Document: %o, %o',
-                                context,
-                                context.document.metadata.document
-                            );*/
+                            // console.log(
+                            //     'Start Document: %o, %o',
+                            //     context,
+                            //     context.document.metadata.document
+                            // );
```
chrisvire, line 388: keep the commented out console.log?

src/lib/components/ScriptureViewSofria.svelte
```diff
@@ -441,12 +448,9 @@ TODO:
                                     workspace.paragraphDiv = document.createElement('div');
                                     workspace.paragraphDiv.classList.add(paraClass);
                                 } else if (sequenceType == 'introduction') {
+                                    workspace.phraseDiv = startPhrase(workspace, 'keep');
```
chrisvire, line 451: What does this do?

src/lib/components/ScriptureViewSofria.svelte
davidmoore1: Yes

src/lib/components/ScriptureViewSofria.svelte
davidmoore1: Primarily, the change moved this to before creating a new paragraph div.  Didn't test this out enough with introductions.  Overall, what the start phrase does is move any existing phrase i (containing the last phrase in the previous text message) into the paragraph div.  The keep means we don't change the phrase index (the a b c d) which isn't needed for introductions.  Prior to the change, the introduction text was showing up in the paragraph following the one it should have.  Most things were displaced by one paragraph which caused some crazy formatting in the glossary.

src/lib/components/ScriptureViewSofria.svelte
davidmoore1: Changing it from a /* */ type to one that the command / will take out, now that I know about that command.  I am uncommenting all of these every time I go to do some change to understand what is coming in and out.

chrisvire (approved):
👍

src/lib/components/ScriptureViewSofria.svelte
davidmoore1: Taking this one out.

PR #181, opened 2023-04-14
Feature: video

chrisvire:
* implement video placement
* support before, after, top, and bottom placement
* for now, include `js/app-builder-video.js` in the repo but later it will be generated by SAB


src/lib/components/ScriptureViewSofria.svelte
```diff
@@ -941,8 +992,21 @@ TODO:
...
+                                const video = config.videos.find((x) => x.id === id);
+                                const divVideoBlock = createVideoBlock(
+                                    document,
+                                    video,
+                                    workspace.currentVideoIndex++
+                                );
+
+                                // TODO: Add it where we currently are in the document?
+                                // This isn't correct, but at least gets it into the document for now
+                                workspace.root.appendChild(divVideoBlock);
```
chrisvire, line 1008: @davidmoore1 I know this is not correct, but it at least gets the videos in the document.  How would I get them put in the current position in the document?

In the video books, the videos are separated by `\b` tags.
In a scripture book, someone could put it after a verse?

PR #182, opened 2023-04-14
Theme Card Icons

Jason-Lopez8:


chrisvire (commented):
I think this might be better fixed with another change I am working.  It turns out that the sab-app.css that is exported during the "Build PWA Data" is hard-coded to the Normal theme. I am working on including all the themes with CSS Variables.  So I don't think this will be required.

chrisvire:
All fixed in #199

PR #194, opened 2023-04-20
Use CSS Vars for Colors

chrisvire:
Scripture App Builder has been changed to generate sab-app.css with CSS vars for colors. Also, clean up where items are added to the `<head>`

The example_data folder needed to be updated for new CSS and I am not sure what the previous project was used. So I updated the project and included a README with the Google Drive location.


static/override-sab.css
```diff
@@ -0,0 +1,4 @@
+body {
```
chrisvire, line 1: SAB stopped excluding margin-top, margin-bottom which was a bit of a hack.  We should exclude it from inside the PWA.

chrisvire:
@FyreByrd Could you look at the main README.md?

FyreByrd (changes requested):
There is one line in `README.md` that needs to be updated to reflect the changes in `package.json`, other than that, this looks fine.

FyreByrd (approved):
Looks good to me

davidmoore1 (approved):

PR #214, opened 2023-05-01
Start new phrase after footnote following separator

davidmoore1:


chrisvire (approved):
👍 

Worked with Matt 1:1 on WEB with footnote at comma

PR #217, opened 2023-05-02
Ignore Strong references

chrisvire:

PR #218, opened 2023-05-02
Ignore /w wrappers for Strong's concordance info

davidmoore1:


chrisvire:
Look at #217. I think it would be better to remove it from the data included in PKF file instead of ignoring it in the render.

chrisvire:
Use #217 instead

Issue #225, opened 2023-05-13
Back/forward arrows on desktop - to move between chapters


rpmargetts:
Since we can't swipe left and right very easily on desktop browsers, it would be good to have a back arrow in the left margin area and a forward arrow in the right margin area - to make it easier to see how to move between chapters.

See for example: https://myk.global.bible/
or https://www.bible.com/bible/116/GEN.2.NLT

PR #236, opened 2023-05-19
Back/forward arrows to move between chapters (#225) 

josieang:
I think we can consider hiding these buttons on small screens, I've left that as a TODO.

The prev button is hidden for the first chapter, and likewise for the next button for the last chapter of the last book.

Small Screens:
image

Larger screens:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,7 @@ A template for building a Progressive Web App with [Scripture App Builder](https
 
 ## Usage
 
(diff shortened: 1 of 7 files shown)


josieang:
It looks like i accidentally touched some of the commits on main and pushed those. I've created a new pull request at https://github.com/sillsdev/appbuilder-pwa/pull/237

PR #246, opened 2023-06-05
Added `changeSpToRem()` in `scripts\convertConfig.ts` to convert from…

Caleb-Parks:
… sp to rem

Wrote a function to convert from sp (android style) to rem, named `changeSpToRem()` in `scripts\convertConfig.ts`.


chrisvire (approved):
👍

PR #273, opened 2023-06-12
Grid View hover

jakedcolbert:


chrisvire:
I looked at the Android code, and it is using "ButtonSelectedColor" when the user is mouseDown over a tile.

chrisvire:
This would be the way that you would have reactive hover color defined from the theme colors:

```
$: buttonSelectedColor = $themeColors['ButtonSelectedColor'];
```

chrisvire (changes requested):
Please use:

`$: buttonSelectedColor = $themeColors['ButtonSelectedColor'];`

jakedcolbert:
Tried to implement and did not work. Used on:touch and on:mouseover to try to replicate hover which is recognized by both mobile and web but does not work either despite touch/mouse events being registered. Needs alternative solution. Somehow bg color is hard stuck and isn't reactive.

jakedcolbert:
Recognized the problem with the new solution:

```html
<span
    on:click={() => handleClick(cell.id)}
    class="dy-btn dy-btn-square dy-btn-ghost normal-case truncate text-clip"
    style={cellStyle}
    style:background-color={hovered === cell ? 'red' : 'blue'}
>
    {cell.label}
</span>
```

hovered is an event target/span being compared to a cell object.

chrisvire (changes requested):
When trying on mobile, the color stays on the first span that  touch and doesn't move and I move to other spans.

jakedcolbert:
No touchenter and touchleave events exist. Only touchstart, touchend, touchmove, and touchcancel are availible.

jakedcolbert:
Here is a solution but on my testing it responds slowly

src/lib/components/SelectGrid.svelte
```diff
@@ -4,11 +4,42 @@ A component to display menu options in a grid.
...
     import config from '$lib/data/config';
     export let options: App.GridGroup[] = [];
     export let cols = 6;
 
+    let hovered = null;
+    $: hoverColor = $themeColors['ButtonSelectedColor'];
+
+    // Function to handle span touch
+    function handleHover(event) {
+        console.log('touched');
```
chrisvire, line 17: Remove debug log

src/lib/components/SelectGrid.svelte
```diff
@@ -4,11 +4,42 @@ A component to display menu options in a grid.
...
+        hovered = event.target.id;
+    }
+
+    // Function to handle span touch end
+    function handleHoverEnd() {
+        console.log('untouched');
```
chrisvire, line 23: Remove debug log

chrisvire (changes requested):
Minor changes

chrisvire:
@jakedcolbert It seems to solve the problem.  However, there are issues with moving the touch around with the app in general.  I will try to address that.

chrisvire:
I fixed the issues.

PR #279, opened 2023-06-13
Implement Annotations using IndexedDB

davidbbryant:
Functional version of bookmarks using IDBDatabase instead of local storage. Should be pretty good overall. A bit concerned about unused variables like docset for $bookmarks but maybe that is a non-issue if they were never really called/needed for outside functions


chrisvire (approved):
👍

PR #306, opened 2023-06-28
active tab

jakedcolbert:
It is not necessary for components to pass in a hardcoded active tab attribute. Neither Book or Chapter selector make use of it, and it makes things complicated for CollectionSelector. If the active tab is set to 'Single Pane' but the 'Single Pane menu is disabled, it will still display the contents since all tabs are being passed to the tab menu despite their visibility:

`showSinglePane: false, showSideBySide: true, showVerseByVerse: true`

Screenshot 2023-06-28 at 3 45 21 PM

It is possible to set the active tab prop to be the 'default layout option' value but Scripture App Builder allows the user to select a disabled layout option as the default. 

Screenshot 2023-06-28 at 3 48 44 PM

Instead, have the tab component menu initialize the active tab to the first available visible tab.

### DIFF
--- a/src/lib/components/BookSelector.svelte
+++ b/src/lib/components/BookSelector.svelte
@@ -192,7 +192,6 @@ The navbar component.
                         visible: showChapterSelector && showVerseSelector
                     }
                 }}
-                active={b}
                 on:menuaction={navigateReference}
             />
         </div>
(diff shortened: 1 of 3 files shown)


jakedcolbert:
@chrisvire

chrisvire:
@jakedcolbert I am fine with using this as the default. The caller should still be able to override the default (i.e. it is an `export let` instead of just `let`).  

With the collection selector, it should set this to the current layout.mode.

chrisvire (approved):
👍

PR #312, opened 2023-07-03
Use bookname in annotation references

davidbbryant:
The reference was set using the bookID which isn't user friendly. Format the reference using the bookName from the book collection.


davidbbryant:
This should be ready for review

chrisvire (approved):
👍

PR #329, opened 2023-07-07
Layout Selector Hotfix

jakedcolbert:
Fixed issue with inconsistency in constant use that caused book selection to be handled incorrectly.


chrisvire (approved):
👍

PR #330, opened 2023-07-10
Chore: Navbar Height #326

### LINKED ISSUE #326: Chore: clean up all pages to use navBarHeight
With #318, we move the Modals and the navBarHeight to `src/routes/+layout.svelte`.  Update all the pages to use the passed in navBarHeight instead of the tailwindcss h-16 class.

jakedcolbert:
Changed navbar height to be assessed by the component itself and not in its call in each page. Navbar height calculated using the `NAVBAR_HEIGHT` store of `view.js`. Resolves #326

### DIFF
--- a/src/lib/components/Navbar.svelte
+++ b/src/lib/components/Navbar.svelte
@@ -10,7 +10,8 @@ The navbar component.
         convertStyle,
         s,
         showDesktopSidebar,
-        direction
+        direction,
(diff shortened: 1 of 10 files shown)


chrisvire (approved):
Good choice to set the hight on the navbar component and get rid of the configuration on the pages.

Issue #252, opened 2023-06-06
Annotation: Sorting


chrisvire:
Dependent on #243 

Implement sorting on the annotations:
Highlights: by reference, by date, by color
Notes: by reference, by date
Bookmarks: by reference, by date

PR #336, opened 2023-07-11
Implement annotation sorting

davidbbryant:
Implement sorting for the bookmarks, highlights, and annotations pages. Pages also now automatically sort by date when opened. Fixes #252

PR #345, opened 2023-07-13
Annotation Sorting

chrisvire:
Try rebasing on main


chrisvire:
Fixes #252. Takes place of #336 to fix issue with linting.

Issue #290, opened 2023-06-22
Layout Configuration: Open on launch


jakedcolbert:
Add support to open the CollectionSelector on launch using `config.mainFeatures["layout-config-first-launch"];`

![Image](https://github.com/sillsdev/appbuilder-pwa/assets/78440533/0eb32b07-ea43-41dc-b41f-0d3c23e250f2)

jakedcolbert:
@chrisvire Is there a distinction between on first launch and every launch?

chrisvire:
@jakedcolbert You can test it out with the Android app.  I would have to as well since I don't know.

jakedcolbert:
@chrisvire After testing on the android app, it only opens on the FIRST launch and does not open at relaunch.

PR #352, opened 2023-07-17
#290 Feature: Layout Menu on First Launch

jakedcolbert:
This pull request implements feature #290 by
- Adding a local storage flag that tracks whether or not this launch is the first ever launch
- Displays modal on first launch based on the following conditionals: 
  - `firstLaunch`: a store object that contains a boolean of whether or not this is the first launch ever on a device.
  - `showColllectionsOnFirstLaunch`: a boolean value pulled from SAB of if the user wants to display the menu on first launch
  - `enoughCollections`: a boolean value assigned to true iff there are at least two book collections for the user to choose from.


chrisvire (approved):
👍

PR #353, opened 2023-07-17
reimplement audio 

### LINKED ISSUE #249: Reimplement Audio
The current audio implementation does not work.

image

1. If audio is playing and one of the other audio buttons is pushed, then the current audio should stop.
2. the skip forward and skip backward on either side of the play button (double triangles) should go back or forward by one verse. These are visible only when timing information is available.
3. The outer skip forward and skip backward (single triangle with line) should go back or forward a section at a time. If there are no sections, then it will go to the previous or next chapter.  [Note: The WEB Gospels content does not have sections. I will find content that does.]
4. If the audio toolbar button is pressed while audio is playing to hide the toolbar, then audio should stop playing.

ChristopherRankin:
fixes #249

### DIFF
--- a/src/lib/components/AudioBar.svelte
+++ b/src/lib/components/AudioBar.svelte
@@ -29,7 +29,6 @@ TODO:
 
     function seekAudio(event) {
         if (!$audioPlayer.loaded) return;
-        // Calculate the percentage of the progress bar that was clicked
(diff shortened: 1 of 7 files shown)


chrisvire (approved):
Great job!

PR #356, opened 2023-07-20
Resolve Layout 'cols' warning

jakedcolbert:
Tab Menu was passing cols value as prop to its child component. Not all child components use the cols property so cols was moved to the props handler of the tab props object that is passed to the child.


chrisvire (approved):
👍

PR #361, opened 2023-07-21
Implement clicking verse number to change audio position

### LINKED ISSUE #251: Audio: while playing audio, tap on verse number changes play location
Depend on #249 being completed first. 

When audio is playing and there is timing information, the user can tap on the verse number to change the playback location.

ChristopherRankin:
* When verse number is clicked audio is played from that verse.
* This fixes #251.
* If the play button or forward/reverse buttons have not been pressed, the verse numbers cannot be clicked.

### DIFF
--- a/src/lib/data/audio.ts
+++ b/src/lib/data/audio.ts
@@ -20,9 +20,6 @@ const cache = new MRUCache<string, AudioPlayer>(10);
 let currentAudioPlayer;
 audioPlayerStore.subscribe(async (value) => {
     currentAudioPlayer = value;
-    if (currentAudioPlayer.audio === null) {
-        return;
-    }
     await getAudio();
 });
 // produces the cache key for the mru audio cache
--- a/src/lib/data/audio.ts
+++ b/src/lib/data/audio.ts
@@ -15,12 +15,14 @@ interface AudioPlayer {
     book: string;
     chapter: string;
     timer: number;
-    isAudio: boolean;
 }
 const cache = new MRUCache<string, AudioPlayer>(10);
 let currentAudioPlayer;
 audioPlayerStore.subscribe(async (value) => {
     currentAudioPlayer = value;
+    if (currentAudioPlayer.audio === null) {
+        return;
+    }
     await getAudio();
 });
 // produces the cache key for the mru audio cache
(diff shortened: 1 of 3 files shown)


src/lib/data/audio.ts
```diff
@@ -274,16 +278,30 @@ export async function getAudioSourceInfo(item: {
...
-//     if (isSelectableText(target)) {
-//         const start = parseFloat(target.start);
-//         this.audio.currentTime = start;
-//         this.audio.play();
-//     } else {
-
-//     }
-// }
+// changes audio to the verse number clicked on
+export function verseClickSoundChange(element) {
```
chrisvire, line 282: The code dealing with the html document should be in ScriptureViewSofira.svelte.  The code in audio.ts should just deal with the audio player.

src/lib/data/audio.ts
```diff
@@ -147,6 +147,10 @@ function updateTime() {
 }
 // calls updateTime() every 100ms
 function toggleTimeRunning() {
+    // enables audio play position to be changed by clicking on verse number
+    if (currentAudioPlayer.isAudio === false) {
```
chrisvire, line 151: It is not good to add additional state which can get out-of-sync.  Could you have a method call hasAudioPlayed which checks the currentAudioPlayer.progress > 0?

chrisvire: If we did keep it as a state variable, "isAudio" is not the best name.

chrisvire (changes requested):
Please move the code dealing with HTML into a function in ScriptureViewSofria.svelte. Have a specific function audio.ts (like `skipToVerse`) that is used.

src/lib/data/audio.ts
```diff
@@ -155,6 +154,14 @@ function toggleTimeRunning() {
     }
     return;
 }
+// checks if audio has played
+function hasAudioPlayed() {
```
chrisvire, line 158: This could simply be:

```js
function hasAudioPlay() { 
    return currentAudioPlayer.progress > 0;
}
```

src/lib/data/audio.ts
```diff
@@ -274,16 +281,28 @@ export async function getAudioSourceInfo(item: {
...
-//     if (isSelectableText(target)) {
-//         const start = parseFloat(target.start);
-//         this.audio.currentTime = start;
-//         this.audio.play();
-//     } else {
-
-//     }
-// }
+// changes audio to the verse number clicked on
+export function verseClickSoundChange(verseId) {
```
chrisvire, line 285: This function is similar to `seek(position)`, right? You just have to find the position from the timing files.

Please change this to:
1. `seekToVerse(verseId)`
2. return if`!hasAudioPlayed()`
3. find the new position from the timing info.
4. call `seek(position)` so this code is not repeated (which will do the pause/play logic).

chrisvire (changes requested):
Some minor changes.

chrisvire (approved):
:+1:

Issue #307, opened 2023-06-30
Annotations: format text without verse numbers


chrisvire:
Verse numbers and footnotes should not be included.

PR #365, opened 2023-07-25
Selected Verses: get raw text for verses

chrisvire:
* Use proskomma query to get raw text of verse
* Delay getting text of selected verses until needed


davidmoore1 (approved):

chrisvire:
Fixes #307

PR #367, opened 2023-07-25
Implement Share All

### LINKED ISSUE #358: Annotation: Share all
From the app bar, there is a share button to share all of the annotations as a txt file. Use Web Share API.

https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API

There is also configuration for the default To: field if sharing by email.

Image

davidbbryant:
This adds the button for and implements the share all functionality for bookmarks, highlights, and notes. It also edits the regular Share functionality by changing it to a function from a separate share file. Has strange error on the bookmarks page which is being investigated. Fixes #358

### DIFF
--- a/src/lib/data/annotation-share.ts
+++ b/src/lib/data/annotation-share.ts
@@ -1,0 +1,34 @@
+import config from '$lib/data/config';
+
+export async function shareAnnotation(annotation: any) {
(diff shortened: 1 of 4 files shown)


chrisvire (approved):
👍

PR #370, opened 2023-07-26
Add some margin between page buttons and text

chrisvire:


src/routes/+page.svelte
```diff
@@ -350,7 +350,7 @@
             <div class="hidden md:flex basis-1/12 justify-end">
```
jakedcolbert, line 350: Take a look at these two links. Adding more [margin](https://225-margin.surge.sh) like Richard suggested causes the arrows to collide with the border on almost md breakpoint resolution sizes. Using [justify-center](https://225-justify-center.surge.sh) here and removing the inline margins allows the arrows to exist comfortably in all resolutions.

chrisvire:
Implement using #371

Issue #308, opened 2023-06-30
Annotiations: align text with book name


chrisvire:

chrisvire:
![Image](https://github.com/sillsdev/appbuilder-pwa/assets/517814/9467127a-f465-4b21-b3b0-1d3b2d2c857b)

PR #373, opened 2023-07-26
Implement Text Alignment

davidbbryant:
Implements proper text alignment of the main body text of the bookmarks, highlights and notes cards so that it starts after the icon and does not go past the three dot dropdown menu. Fixes issue #308


chrisvire (approved):
👍

PR #375, opened 2023-07-26
Implement Padding

### LINKED ISSUE #332: Annotations: Add Padding
Each of the annotations pages (bookmarks, highlights, notes) needs to have padding between the top navbar and the first data card. The padding should look similar to that between the different cards.

davidbbryant:
Added padding inside the card holding div in bookmarks, highlights, history, and notes in order to force proper spacing of the cards from the sides of the page and the top bar. Fixes #332


chrisvire (approved):
👍

PR #390, opened 2023-08-03
Bug: Footnote Caller Incorrectly Styled

jakedcolbert:
Footnote class in root span styles footnote caller incorrectly.

Resolve by removing footnote class from root span.


chrisvire (approved):
👍

Issue #322, opened 2023-07-05
UI: Style LayoutOptions menu


jakedcolbert:
Style changes to the CollectionSelector that supports reactivity for the following theme options

<img width="723" alt="Screenshot 2023-07-05 at 3 47 42 PM" src="https://github.com/sillsdev/appbuilder-pwa/assets/78440533/59a7850b-e67c-4817-85fa-8fc17ec40c2d">

jakedcolbert:
![Image](https://github.com/sillsdev/appbuilder-pwa/assets/78440533/6d81d484-7297-4164-bb21-97f19b946b19)

jakedcolbert:
First tab should be active when opening the modal

jakedcolbert:
Text Styles and UI Styles

PR #398, opened 2023-08-16
Fix CollectionSelector

chrisvire:
* dy-menu-compact removed in DaisyUI v2; replace with dy-menu-sm
* remove unused vertOffset

### DIFF
--- a/src/lib/components/CollectionList.svelte
+++ b/src/lib/components/CollectionList.svelte
@@ -19,7 +19,7 @@ Custom list of collections for the LayoutOptions menu
     }
 </script>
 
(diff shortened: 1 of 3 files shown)


chrisvire:
@jakedcolbert I assume that we don't need the vertOffset stuff anymore since we are centering the modal instead of trying to position it below the appbar (like the TextAppearanceSelector).

jakedcolbert:
LGTM. Daisy UI menu will likely not be used with #322 but is good for now.

jakedcolbert (approved):

chrisvire:
Looks like this is not needed.

PR #420, opened 2023-11-28
Figure graft crashes page display

davidmoore1:
At this point, ignore figure graft.  At some point need to implement displaying the figure, but for now, this change prevents a figure graft from causing an exception that prevents any page with a figure graft from displaying at all


chrisvire (approved):
👍

PR #431, opened 2023-12-07
Fix crash on book selector

davidmoore1:
Problem can occur if you are going from a high chapter number (Like Genesis 50) to a book with fewer chapters.  The initial setup of the verse grid is using the old chapter number and not finding it in the new book, which causes an exception in SelectGrid.


davidmoore1:
If switching books, set the nextRef chapter to point to 1 initially

PR #432, opened 2024-01-12
Add support for markdowns 

davidmoore1:


src/lib/scripts/markdown.ts
chrisvire, line 1: Should the filename be changed since it isn't markdown anymore?

scripts/convertMarkdown.ts
```diff
@@ -0,0 +1,320 @@
...
+    const refLower = ref.toLowerCase();
+    return refLower.startsWith('tel:');
+}
+function isLocalAudioFile(ref: string): boolean {
+    let result = false;
+    const refLower = ref.toLowerCase();
+    if (!refLower.startsWith('http')) {
+        const ext = getFilenameExt(refLower);
+        if (ext != null) {
+            result = ext === 'mp3' || ext === 'ogg' || ext === 'wav';
```
chrisvire, line 85: Does the Android code have .ogg?  I would have thought it would be .webm.

chrisvire: Actually, for the web, Safari can't play .ogg or .webm files. So this should probably be just mp3 and wav.

chrisvire (changes requested):
Minor change with audio

scripts/convertMarkdown.ts
davidmoore1: That is the way it is coded in BaseWriter.java.  I made the change to take that out of the pwa code.

davidmoore1: Changed it slightly to acknowledge webm and ogg links as audio links, but then to ignore them, not producing a link html since the file type is not supported

chrisvire (approved):
👍

PR #445, opened 2024-03-15
Handle references with verse ranges and other exceptions

davidmoore1:


chrisvire (approved):
👍 

We will need some formatting until we render a portion using ScriptureViewSofria

PR #447, opened 2024-03-21
Apply text font setup to footnote

davidmoore1:
Apply the same font, font size and line height attributes to the footnote as are applied to the standard text.  Makes the reference part bold and the rest the correct height and size.


chrisvire:
Example USFM from David:
https://drive.google.com/file/d/1XVcG6lsNmxxCN9ezdd6FgJYQYhaONh_s/view?usp=drive_link

Issue #455, opened 2024-04-05
Chapter footers appearing where they shouldn't


davidmoore1:
I have an app with 4 book collection.  The first collection has a footer defined.  The others do not.  The footer appears on all book collections instead of just the first.

PR #463, opened 2024-04-12
Footnotes display on all collections after it displays once (Issue #455)

davidmoore1:
Issue #455


davidmoore1:
Original footnote was being added to the container rather than the workspace.root.  The container is never cleared.  As a result, once the footnote displayed for a book collection for which the figure was configured, it stayed there when the book collection changed.

chrisvire (approved):
👍

Issue #462, opened 2024-04-12
Figure without a trailing paragraph crashes pwa


davidmoore1:
There is an issue with Proskomma that if a figure exists without a trailing paragraph, it gets confused and verse text following the figure comes in as a usfm:f type.  Our parsing code is trying to process this as footnote text and crashes as a result.

benpaj-cps:
Closed by #464.

PR #464, opened 2024-04-12
Avoid crash on figure without trailing paragraph (Issue #462)

davidmoore1:
Issue #462


davidmoore1:
This performs a check to make sure that the proper sequence is set when the text for the footnote or xref is processed.  It also does a check to make sure that the footnoteDiv that is being added to exists for safety.  A check for 'fig' that was in the method was deleted as it was a leftover check from before figures were handled that was not removed when those changes were made.

chrisvire (approved):
👍

benpaj-cps:
Closes #462.

PR #465, opened 2024-04-16
During convertBooks, remove figs with missing illustrations

chrisvire:
* For many Bibles, there are illustrations for the print bible and the related`\fig` markers are left in the USFM that are not used in the mobile app. The native app ignores the figs referencing missing illustrations.

### DIFF
--- a/scripts/convertBooks.ts
+++ b/scripts/convertBooks.ts
@@ -32,11 +32,48 @@ function removeStrongNumberReferences(text: string, _bcId: string, _bookId: stri
     return text.replace(/(\\\+?w) ([^|]*)\|strong="[^"]*"\1\*/g, '$2');
 }
 
+function removeMissingFigures(text: string, _bcId: string, _bookId: string): string {
+    // Regular expression to match \fig markers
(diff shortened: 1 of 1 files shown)


scripts/convertBooks.ts
```diff
@@ -32,11 +32,48 @@ function removeStrongNumberReferences(text: string, _bcId: string, _bookId: stri
...
+    const figRegex = /\\fig\s(.*?)\\fig\*/g;
+
+    // Replace each \fig marker with the appropriate action
+    return text.replace(figRegex, (match, figContent) => {
+        // Split the content of the \fig marker by pipe (|)
+        const parts = figContent.split('|');
+
+        // Extract the image source from the first part
+        let imageSource;
+        if (parts[1].startsWith('src="')) {
```
davidmoore1, line 46: Does it have to start with this string, or should this string just be included somewhere if it is the new format?

chrisvire: Hmm ... I guess it technically doesn't have to start with it.

davidmoore1 (approved):

PR #466, opened 2024-04-16
Fix setting verse-numbers setting

chrisvire:
Without the initial value set to an empty map, the array.reduce didn't correctly assign the key,defaultValue for verse-numbers,


davidmoore1 (commented):
LGTM

Issue #482, opened 2024-04-29
show border book feature is ignored


davidmoore1:
Show border can be set by book.  The feature is not being used by PWA

PR #483, opened 2024-04-29
issue #482 - Show border book feature is being ignored

davidmoore1:


src/routes/+page.svelte
```diff
@@ -131,7 +131,12 @@
     const showCollectionsOnFirstLaunch = config.mainFeatures['layout-config-first-launch'];
     const showCollectionViewer = config.mainFeatures['layout-config-change-viewer-button'];
     const showAudio = config.mainFeatures['audio-allow-turn-on-off'];
-    $: showBorder = config.traits['has-borders'] && $userSettings['show-border'];
+    $: showBorderSetting = getFeatureValueString(
```
chrisvire, line 134: Why use `getFeatureValueString` instead `getFeatureValueBoolean`?  It seems that the values are `true`, `false`, or `inherited`.

src/routes/+page.svelte
```diff
@@ -131,7 +131,12 @@
...
+            'show-border',
+            $refs.collection,
+            $refs.book
+        );
+    $: showBorder = config.traits['has-borders'] && $userSettings['show-border'] && (showBorderSetting !== 'no');
```
chrisvire, line 139: Why compare against `no` ?

chrisvire (approved):
👍

Issue #471, opened 2024-04-18
Fullscreen


chrisvire:
With the native app, the user can double-tap on the screen and go into full-screen mode.

There is a feature to control whether this feature is available:

```
    <feature name="full-screen-double-tap" value="true"/>
```

The feature is controlled in the UI from the Appearance > Formatting > Text Size page.

![image](https://github.com/sillsdev/appbuilder-pwa/assets/517814/22d5498d-faa9-4b17-a476-2b99ba1d5cc1)

jakedcolbert:
Visual experience notes: 

- User double taps the screen to initiate full-screen
- On full-screen, an alert displays from the top bar explaining to the user that they can exit full screen by swiping from the top bar.
- The user can exit full-screen by swiping down from the top bar or by double-tapping the screen again. 
- In full-screen, all elements besides the scripture view are hidden from the user.

chrisvire:
Are these based on using the Android app?

chrisvire:
I don't see this message on newer versions of Android.  So I wouldn't implement it right now.

jakedcolbert:
@chrisvire Understood. Then the requirements are as follows: 

- User double taps the screen to initiate full-screen
- The user can exit full-screen by swiping down from the top bar or by double-tapping the screen again. 
- In full-screen, all elements besides the scripture view are hidden from the user.

chrisvire:
When I swipe down from the top of the screen, I get an overlay from Android.  It doesn't exit full-screen.

jakedcolbert:
@chrisvire Understood part 2! Then the requirements are as follows: 

- User double taps the screen to initiate full-screen
- The user can exit full-screen by double-tapping the screen again. 
- In full-screen, all elements besides the scripture view are hidden from the user.

chrisvire:
Sounds good for right now.

I sent an email to Richard Margetts (the product owner and main developer of the Android app) to clarify some interaction between the full-screen mode and the audio bar.  I CC'ed you (your gordon email) on the message.

chrisvire:
Sending another message to ask Richard his opinion.  Here was my original message:

<img width="852" height="736" alt="Image" src="https://github.com/user-attachments/assets/c423c697-f4bd-4caa-a8af-e581164c4a65" />

PR #485, opened 2024-04-30
Feature: Fullscreen #471

jakedcolbert:
Implements #471


chrisvire (changes requested):
When I use your PR, I see these issues:
* the audio toolbar is not stuck at the bottom of the window
* the border isn't just around the window
* I see the wait loading animation over the top of the text when I scroll down from the first page of text

I am using it on the device
connected to the http://{network}:5173/ server

PR #519, opened 2024-05-09
Detect System Language of Browser

### LINKED ISSUE #487: Detect system language of browser
on the Appearance > Interface > Languages page in SAB, there are preferences for which language to use:

image

These settings show up in the converted `config.js`:

```
  "interfaceLanguages": {
    "useSystemLanguage": true,
    "writingSystems": {
      "en": {
        "fontFamily": "system",
        "textDirection": "LTR",
        "displayNames": {
          "en": "English",
          "zh": "英文"
        }
      },
      "zh": {
        "fontFamily": "system",
        "textDirection": "LTR",
        "displayNames": {
          "en": "Chinese (Simplified)",
          "zh": "中文（简体)"
        }
      },
      "bo": {
        "fontFamily": "font1",
        "textDirection": "LTR",
        "displayNames": {
          "en": "Tibetan"
        }
      }
    }
  },
"translationMappings": {
    "defaultLang": "en",
```

The `useSystemLanguage` property in the `interfaceLanguages` object indicates if they have checked the first item in the dialog.  The `defaultLang` property in the `translationMappings` object specifies the value of the dropdown in either the first or second item in the dialog.

in `src/lib/data/stores/localization.js` we are already getting the `languageDefault` but we are not using the `useSystemLanguage` setting.

To get the language of the browser, use `window.navigator.language`.  There will be some work to determine if that matches one of the languages.

gmjgeek:
Fixes #487 

On the Appearance > Interface > Languages page in SAB, there are preferences for which language to use:

image

If the app is configured to use the system language as default, it obtains preferred languages from the browser and compares them to available UI translations.  If a matching translation is found (or a similar one, ex: en-US instead of en-UK), this is the default language.  Otherwise, the app will use the fallback language specified in SAB.

The logic for choosing a language that matches the user's preference is found in `getDefaultLanguage()` in `src/lib/data/language.ts`.

The following screenshots demonstrate how the "Preferred Language" setting in Google Chrome affects the default language in the app on a mobile device.

image

image

### DIFF
--- a/src/lib/data/stores/localization.js
+++ b/src/lib/data/stores/localization.js
@@ -1,10 +1,11 @@
 import { derived } from 'svelte/store';
 import { userSettings } from './setting';
+import { getLanguages } from '$lib/data/language';
 import config from '../config';
 
 /** localization */
-export const languageDefault = config.translationMappings.defaultLang;
-export const languages = Object.keys(config.interfaceLanguages.writingSystems);
+export const languageDefault = userSettings['interface-language']?.defaultValue;
+export const languages = getLanguages();
 export const language = derived(
     userSettings,
     ($userSettings) => $userSettings['interface-language']
--- a/src/lib/data/stores/localization.js
+++ b/src/lib/data/stores/localization.js
@@ -4,8 +4,10 @@ import { getLanguages } from '$lib/data/language';
 import config from '../config';
 
 /** localization */
-export const languageDefault =
-    userSettings['interface-language']?.defaultValue ?? config.translationMappings.defaultLang;
+
+// If a word can't be translated in the current language, use languageDefault.
+export const languageDefault = config.translationMappings.defaultLang;
+
 export const languages = getLanguages();
 export const language = derived(
     userSettings,
--- a/src/lib/data/stores/localization.js
+++ b/src/lib/data/stores/localization.js
@@ -4,7 +4,8 @@ import { getLanguages } from '$lib/data/language';
 import config from '../config';
 
 /** localization */
-export const languageDefault = userSettings['interface-language']?.defaultValue;
+export const languageDefault =
+    userSettings['interface-language']?.defaultValue ?? config.translationMappings.defaultLang;
 export const languages = getLanguages();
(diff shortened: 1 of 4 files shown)


src/lib/data/stores/localization.js
```diff
@@ -1,10 +1,11 @@
 import { derived } from 'svelte/store';
 import { userSettings } from './setting';
+import { getLanguages } from '$lib/data/language';
 import config from '../config';
 
 /** localization */
-export const languageDefault = config.translationMappings.defaultLang;
-export const languages = Object.keys(config.interfaceLanguages.writingSystems);
+export const languageDefault = userSettings['interface-language']?.defaultValue;
```
chrisvire, line 7: This won't work if the UserSetting for interface-language is not made available to the end user.

chrisvire (changes requested):
You will need to handle when the user is not offered a choice of languages and should be using the defaultLang.

How do handle choosing a language that matches the language of the browser?

gmjgeek:
The function `getDefaultLanguag()` handles choosing a language that matches the language of the browser.  This function is in `src/lib/data/language.ts`.

src/lib/data/stores/localization.js
```diff
@@ -1,10 +1,12 @@
 import { derived } from 'svelte/store';
 import { userSettings } from './setting';
+import { getLanguages } from '$lib/data/language';
 import config from '../config';
 
 /** localization */
-export const languageDefault = config.translationMappings.defaultLang;
-export const languages = Object.keys(config.interfaceLanguages.writingSystems);
+export const languageDefault =
```
chrisvire, line 7: languageDefault is just an alias for `config.translationMappings.defaultLang`.  

If you look at the current usage in SideBar.svelte, it is used to find the fallback translation in case the MenuItem.title doesn't have a translation based on the current language.

chrisvire (changes requested):
Please revert the change to languageDefault.

chrisvire (approved):
Very good work!

Issue #516, opened 2024-05-09
Footer: Use book defined footer over collection footer if exists


chrisvire:
Each book collection and each book has a configuration for the footer for the chapter.

![image](https://github.com/sillsdev/appbuilder-pwa/assets/517814/cab1831e-2c32-4257-96e4-490bba8bb979)

Each book can override his by first turning on "use book-specific setting" on  Book > Features page.

![image](https://github.com/sillsdev/appbuilder-pwa/assets/517814/de8d2e7a-01b6-4c20-95c5-6f1695b1fd9c)

This displays a Footer tab on the book page

![image](https://github.com/sillsdev/appbuilder-pwa/assets/517814/ce495594-5f7d-4abb-88b1-e5ecde634963)

In `ScriptureViewSofria.svelte`, only the book collection footer is currently used in `addFooter`.  That code should check if the current book has a footer definition.

PR #527, opened 2024-05-10
Footer: Display Markdown as HTML

### LINKED ISSUE #515: Footer: Convert Markdown in footer configuration
Each book collection and each book has a configuration for the footer for the chapter.

image

Each book can override his by first turning on "use book-specific setting" on  Book > Features page.

image

This displays a Footer tab on the book page

image

Scripture App Builder allows markdown to exist in book content which is not USFM markers. During the `convert` process, we transform the markdown to USFM milestones so that they can be parsed by Proskomma, the USFM parser and then processed by `ScriptureViewSofria.svelte`.

These footers are in `config.js` and we want to convert the markdown to HTML so that they can be inserted by `ScriptureViewSophria.svelte` in the function `addFooter` which currently adds the contents as innerHTML to a span.

There is code in `scripts/convertMarkdown.ts` that converts markdown to the USFM milestones and used by `scripts/convertBooks.ts`.  Maybe it can be refactored so that it can be used by `scripts/convertConfig.ts` where it reads in the footer properties?

gmjgeek:
Fixes #515

During the conversion process, SAB converts supported Markdown in footers to HTML.

The footers for the Book Collection are rendered at the bottom of each page, as the screenshot below shows.  However, as issue #516 explains, the PWA currently ignores book-specific footers.

image

image

### DIFF
--- a/scripts/convertConfig.ts
+++ b/scripts/convertConfig.ts
@@ -2,6 +2,7 @@ import { readFileSync, existsSync, PathLike, readdirSync } from 'fs';
 import jsdom from 'jsdom';
 import path from 'path';
 import { Task, TaskOutput } from './Task';
+import { convertMarkdownsToHTML } from './convertMarkdown';
 
 /**
  * TODO:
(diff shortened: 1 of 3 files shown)


chrisvire (approved):
👍

PR #532, opened 2024-05-13
Catalog as Data

### LINKED ISSUE #497: Split the catalog.js per collection and treat as data instead of project code
The `src/lib/data/catalog.js` is generated during the `convert` process and contains a index of all of the collection / book / chapter / verse information so that navigation doesn't have to parse the scripture data to get this information.  This is fine for 1 or 2 collections.  If you have more (like 8), then the file gets really large, the project build time gets long (on my M1 Max Mac Studio, it took 28 minutes), and the load time for the app is very long. 

We should split up the conversion of the catalog by collection and store with the data and load it like we load the proskomma freeze files (pkf) on demand.

gmjgeek:
Fixes #497

Builds the catalog as a collection of static JSON files.  Each book collection has its own catalog file, which is fetched at runtime when the user changes book collections. 

Since the catalog is no longer a source file, build times are much faster.  For example, a project that used to take ~20 minutes to build now builds in ~30 seconds (~1 min 30 sec including the conversion process).

In testing, changing book collections took about 0.5 seconds on a mobile device.  The app was built in release mode and was running over a local network.

To test performance in slow network conditions, the app was accessed on a PC via Google Chrome with network throttling set to "Slow 3G." Under these conditions, the app took a little over a second to change book collections.

As an experiment, the implementation was modified to generate the catalog dynamically instead of loading static catalog files (see gmjgeek/appbuilder-pwa@e32cac07f5b54e6b239e22f9448cadf9ec4be136).  With this implementation, the app took 3-4 seconds to change book collections on a mobile device.  For this reason, it seems preferable to fetch the catalog from the server.


chrisvire (changes requested):
1) Fix initialization of ReferenceStore to pull value from localStorage.refs
2) For fontChoices, return `[]` instead of `undefined` when $refs is not initialized.

I am looking forward to passing this change along to several projects that are blocked due to number of book collections.

chrisvire (approved):
@gmjgeek This is excellent work.  Great job!

PR #540, opened 2024-05-15
display message when there is no history

### LINKED ISSUE #503: Empty History should display message
If the history is empty (which you can create by pressing the trash can in the history display window) a message should be displayed saying there is no history.
Simulator Screenshot - iPhone 12 - 2024-05-08 at 13 56 45

mikaylahubbard:
fixes #503.
when there is no history, display the message to indicate that. This way it doesn't look like an error.

### DIFF
--- a/src/routes/history/+page.svelte
+++ b/src/routes/history/+page.svelte
@@ -30,6 +30,10 @@
     </div>
 
     <div class="overflow-y-auto p-2.5">
+        {#if data.history.length === 0}
+            <div class="history-message-none">{$t['History_None']}</div>
+            <div class="history-message-none-info">{$t['History_None_Info']}</div>
+        {/if}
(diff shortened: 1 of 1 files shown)


src/routes/history/+page.svelte
```diff
@@ -30,6 +30,10 @@
     </div>
 
     <div class="overflow-y-auto p-2.5">
+        {#if data.history.length === 0}
+            <div class="history-message-none">{$t['History_None']}</div>
+            <div class="history-message-none-info">{$t['History_None_Info']}</div>
+        {/if}
```
chrisvire, line 36: Please put the iteration of the data.history items in an else branch of the if condition

chrisvire (changes requested):
Minor change.

chrisvire (approved):
👍

PR #544, opened 2024-05-15
Display message when there are no annotations

### LINKED ISSUE #541: Empty Annotations should display message
Just like history displays a message when there are no entries (#503), the same thing happens with annotations (bookmarks, notes, and highlights).

image

mikaylahubbard:
Fixes #541

Works the same as "display message when there is no history". For the bookmarks, highlights, and notes,
when there is no annotation, display a message.
This way it does not look like an error.

### DIFF
--- a/src/routes/bookmarks/+page.svelte
+++ b/src/routes/bookmarks/+page.svelte
@@ -4,7 +4,7 @@
     import { BookmarkIcon } from '$lib/icons';
     import ShareIcon from '$lib/icons/ShareIcon.svelte';
     import Navbar from '$lib/components/Navbar.svelte';
-    import { bookmarks, refs, t } from '$lib/data/stores';
+    import { refs, t } from '$lib/data/stores';
     import { formatDate } from '$lib/scripts/dateUtils';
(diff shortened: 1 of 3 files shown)


src/routes/bookmarks/+page.svelte
```diff
@@ -4,7 +4,7 @@
     import { BookmarkIcon } from '$lib/icons';
     import ShareIcon from '$lib/icons/ShareIcon.svelte';
     import Navbar from '$lib/components/Navbar.svelte';
-    import { refs, t } from '$lib/data/stores';
+    import { bookmarks, refs, t } from '$lib/data/stores';
```
chrisvire, line 7: `bookmarks` was accidentally added.  Please remove.

chrisvire (changes requested):
Minor change.

src/routes/bookmarks/+page.svelte
chrisvire: @mikaylahubbard I was able to change this through the website.  I would like to get this into the release that will be happening today or tomorrow

chrisvire (approved):
👍

PR #550, opened 2024-05-18
Fix issues with badges

chrisvire:
* Get the languages from convert process instead of hardcoded list
* Try to use interface language for Google Play link, but fallback to languageDefault and then 'en'
* Fallback to languageDefault for App Store image, then to english
* Make sure en App Store image is included for fallback


aidanpscott (commented):
Looks good.

aidanpscott (commented):
also looks good.

PR #555, opened 2024-05-20
Feature/data sandbox 523

### LINKED ISSUE #523: Implement data-sandbox
The PWA uses Proskomma to store and query data using GraphQL.  There is a project to browse the data stored in Proskomma.

https://github.com/Proskomma/diegesis-apollo-sandbox

Here are instructions to use getting our data to work with it:
```
mv filename.pkf filename.gz
gunzip filename.gz
replace metata > selectors > [lang, abbr] with [source, project, revision]
node index.js directory
```

graphql_with_mark_howe: 
https://drive.google.com/file/d/1Zm8dliq8nHWbZDTFBu1lwte8h5Lw5VLg/view?usp=drive_link

```
query ($bookCode: String!) {
 docSets {
   document(bookCode: $bookCode) {
     mainBlocks {
       text
       tokens(withScopes: "span/k") {
         payload
       }
     }
   }
 }
}
```

We want to add a data-sandbox folder at the top of the repo that uses the same libraries but gets the data from `static/collections` and chooses the collections from `src/lib/data/config.js`.  It can use `sab-proskomma` and `thaw` from our code.

gmjgeek:
Fixes #523 

Adapts the script from https://github.com/Proskomma/diegesis-apollo-sandbox to allow interactive GraphQL queries on proskomma objects.  To start the sandbox, run the conversion script, then run `npm run sandbox`.  The sandbox will run on localhost at the URL printed to the console.

image

image


chrisvire (changes requested):
data-sandbox should be outside of src directory since it is a utility outside of the PWA (similar to convert). It should not be included in the PWA src tree.

I would suspect that its dependencies would be in the package.json file in the data-sandbox directory (similar to convert).

src/lib/scripts/package.json
```diff
@@ -0,0 +1 @@
+{}
```
chrisvire, line 1: Should this file be deleted?

chrisvire (changes requested):
Looks good!  Just one file to remove, I think.

src/lib/scripts/package.json
gmjgeek: If I remove that file, I get this error when I run `npm run sandbox`: 

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: C:\Users\jones\Code\sil\appbuilder-pwa\src\lib\scripts\thaw.ts

I found that creating a package.json in the scripts folder resolved this issue, even if the JSON was empty.  I wish I could leave a comment about this, but I'm not sure if comments are possible in JSON.

chrisvire: Very odd!

gmjgeek: I just added a "description" field that explains the issue

chrisvire (approved):
This is really good work figuring out to reuse the data-sandbox in our project.  I hope this is useful in working on search and other graphql work in the future.

Issue #560, opened 2024-05-21
ConvertContent crashes with "other" type content entries


davidmoore1:
Content link type "other" entries contain a location field and not a target field (except for "website" location which also has a target).  ConvertContent assumes that target field exists and does not process location.

PR #561, opened 2024-05-21
Issue #560 - Convert Content crashes on "other" link entries

davidmoore1:
Handle link type "other" such as: ```<link type="other" location="layout" />```
Add a linkLocation field to the contents entries and set linkTarget to "" if not defined.
For website entries ( ```<link type="other" location="website" target="https://www.sil.org" /> ```) fill in both fields.

### DIFF
--- a/convert/convertContents.ts
+++ b/convert/convertContents.ts
@@ -103,8 +103,8 @@ export function convertContents(dataDir: string, verbose: number) {
 
             const linkTags = itemTag.getElementsByTagName('link');
             const linkType = linkTags[0]?.attributes.getNamedItem('type')!.value;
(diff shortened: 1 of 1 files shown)


convert/convertContents.ts
```diff
@@ -102,7 +103,8 @@ export function convertContents(dataDir: string, verbose: number) {
 
             const linkTags = itemTag.getElementsByTagName('link');
             const linkType = linkTags[0]?.attributes.getNamedItem('type')!.value;
-            const linkTarget = linkTags[0]?.attributes.getNamedItem('target')!.value;
+            const linkTarget = linkTags[0]?.attributes.getNamedItem('target')?.value ?? '';
```
chrisvire, line 106: assign to undefined instead of empty string?

chrisvire (approved):
👍

PR #563, opened 2024-05-22
contents menu

mikaylahubbard:
links aren't working, there is a lot of work in progress

### DIFF
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -25,6 +25,7 @@
         t,
         themeColors,
         userSettings,
+        contentsStack,
         modal,
         MODAL_TEXT_APPERANCE,
         MODAL_COLLECTION,
@@ -44,12 +45,14 @@
     import StackView from '$lib/components/StackView.svelte';
     import Navbar from '$lib/components/Navbar.svelte';
     import config from '$lib/data/config';
+    import contents from '$lib/data/contents';
     import ScriptureViewSofria from '$lib/components/ScriptureViewSofria.svelte';
     import { getFeatureValueString, getFeatureValueBoolean } from '$lib/scripts/configUtils';
     import { pinch, swipe } from 'svelte-gestures';
     import TextSelectionToolbar from '$lib/components/TextSelectionToolbar.svelte';
     import { base } from '$app/paths';
     import { page } from '$app/stores';
+    import { goto } from '$app/navigation';
     import { onDestroy, onMount, afterUpdate } from 'svelte';
 
     let savedScrollPosition = 0;
@@ -304,11 +307,20 @@
             textCopied = false;
         }, 3000);
     }
+
+    function handleBackNavigation(event) {
+        event.preventDefault();
+        if ($contentsStack.length > 0) {
+            const menuId = contentsStack.popItem();
+            goto(`${base}/contents/${menuId}`);
+        }
+    }
+    $: showBackButton = contents?.features?.['navigation-type'] === 'up' && $contentsStack.length > 0
 </script>
 
 <div class="grid grid-rows-[auto,1fr,auto]" style="height:100vh;height:100dvh;">
     <div class="navbar">
-        <Navbar>
+        <Navbar on:backNavigation={handleBackNavigation} showBackButton={showBackButton}>
             <div
                 slot="left-buttons"
                 class={showOverlowMenu ? 'hidden md:flex flex-nowrap' : 'flex flex-nowrap'}
--- a/src/routes/contents/[id]/+page.svelte
+++ b/src/routes/contents/[id]/+page.svelte
@@ -2,6 +2,7 @@
     import { page } from '$app/stores';
     import Navbar from '$lib/components/Navbar.svelte';
     import {
+        language,
         s,
         t,
         themeColors,
@@ -17,6 +18,8 @@
 
     $: highlightColor = $themeColors['ContentsItemTouchColor'];
     $: title = setTitle($page);
+
+    console.log(config.interfaceLanguages);
 
     function onClick(event, item) {
         event.target.style.background = highlightColor;
@@ -110,10 +113,10 @@
                 title = config.name;
                 break;
             case 'screen':
-                title = page.data.menu.title.default;
+                title = page.data.menu.title[$language] ?? page.data.menu.title.default ?? '';
                 break;
             case 'custom':
-                title = page.data.title.default;
+                title = page.data.title[$language] ?? page.data.title.default ?? '';
                 break;
             case 'none':
             default:
@@ -175,16 +178,16 @@
                         <div class="contents-text-block">
                             <!-- check for title -->
                             {#if $page.data.features['show-titles'] === true}
-                                {#if item.title.default}
-                                    <div class="contents-title">{item.title.default}</div>
-                                {/if}
+                                <div class="contents-title">
+                                    {item.title[$language] ?? item.title.default ?? ''}
+                                </div>
                             {/if}
 
                             <!--Check for subtitle-->
                             {#if $page.data.features['show-subtitles'] === true}
-                                {#if item.subtitle.default}
-                                    <div class="contents-subtitle">{item.subtitle.default}</div>
-                                {/if}
+                                <div class="contents-subtitle">
+                                    {item.subtitle[$language] ?? item.subtitle.default ?? ''}
+                                </div>
                             {/if}
 
                             <!--check for reference -->
--- a/src/routes/contents/[id]/+page.svelte
+++ b/src/routes/contents/[id]/+page.svelte
@@ -118,24 +118,23 @@
             previously used custom title, is this ok?
             i think its fine because you only use that for the custom title option
         */
+       let title = '';
         switch (page.data.features['title-type']) {
             case 'app-name':
                 //where to find the app name: config.js, line 2
-                return config.name;
+                title = config.name;
                 break;
             case 'screen':
-                return page.data.menu.title.default;
+                title = page.data.menu.title.default;
                 break;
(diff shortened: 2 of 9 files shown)


src/routes/+page.svelte
```diff
@@ -304,11 +307,20 @@
             textCopied = false;
         }, 3000);
     }
+
+    function handleBackNavigation(event) {
+        event.preventDefault();
```
chrisvire, line 312: I know I wrote this, but I wonder if the `event.preventDefault()` call should be inside the if condition...

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+        event.target.style.background = highlightColor;
+        setTimeout(() => {
+            clicked(item);
+        }, 100);
+    }
+
+    function checkImageSize(item) {
+        if (item.features['image-width']) {
+            return 'width: ' + item.features['image-width'];
+        } else {
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:31]: You don't need the `else` condition.  Remove it.

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+    async function clicked(item) {
+        //check type of link
+        switch (item.linkType) {
+            //reference linkType
+            case 'reference':
+                contentsStack.pushItem($page.data.menu.id);
+                setReference(item);
+                goto(`${base}/`);
+                break;
+            //need to check if these are the correct strings to use
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:44]: Is this comment needed still?

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+            case 'screen':
+                //not handled yet
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:46]: Seems like it is handled now.  Remove comment? I guess I should have when I implemented it. ;-)

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+                //goes to another contents page
+                contentsStack.pushItem($page.data.menu.id);
+                await goto(`${base}/contents/${item.linkTarget}`);
+                break;
+            case 'other':
+                //not handled yet
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:52]: This one should be removed too.

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+                //switch on item.linkLocation
+                switch (item.linkLocation) {
+                    case 'about':
+                    case 'settings':
+                        goto(`${base}/${item.linkLocation}`);
+                        break;
+                    case 'layout':
+                        modal.open(MODAL_COLLECTION);
+                        //also the title for this one says &amp; instead of & in the json
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:61]: Remove comment?

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+                // For other book types (e.g. quiz), the linkType will be
+                // the book type and the linkLocation will have the route
+                // to the viewer of the book type.
+                goto(`${base}/${item.linkLocation}`);
+                break;
+        }
+    }
+
+    function setReference(item) {
+        // reference = formatReferenceTarget(item.linkTarget)
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:80]: What is this comment for?

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+        /* title options:
+            - app-name 
+            - screen (current contents screen title)
+            - custom 
+            - none
+        
+            I also noticed that if you build files with a custom 
+            title and then again without specifying a title aka 
+            any of the other options, the title in the json was the 
+            previously used custom title, is this ok?
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:122]: It is ok.  You might change the comment to note that the custom title might be in the JSON even though it isn't currently a custom title type.  ... or we could change the convertContents to not pass it along (marginal value in doing this though)

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+            i think its fine because you only use that for the custom title option
+        */
+        let title = '';
+        switch (page.data.features['title-type']) {
+            case 'app-name':
+                //where to find the app name: config.js, line 2
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:128]: What does this comment mean?

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+                                {#if item.title.default === undefined}
+                                    <!--empty-->
+                                {:else}
+                                    <div class="contents-title">{item.title.default}</div>
+                                {/if}
+                            {/if}
+
+                            <!--Check for subtitle-->
+                            {#if $page.data.features['show-subtitles'] === true}
+                                {#if item.subtitle.default === undefined}
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:208]: You can change this so that it a positive check:

`{#if item.subtitle.default }`

and put the div in the if condition and remove the else.

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,229 @@
...
+                                    <!--empty-->
+                                {:else}
+                                    <div class="contents-subtitle">{item.subtitle.default}</div>
+                                {/if}
+                            {/if}
+
+                            <!--check for reference - should this go before the subtitle?-->
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:215]: Clean up comment...

chrisvire (changes requested):
I like that you include comments in the code as you were working on it.  Just need to clean them up so that the comments that remain are useful for someone who will maintain the code.

@davidmoore1 Could you review this as well?

davidmoore1:
I will review the code.  The functional issues I see are not big ones but:
Using PWA test, the link to Mateo, which should set the reference to the book of Matthew does that, but it doesn't adjust the chapter so you end up in Matthew at whatever chapter you were in prior to getting into the contents menu to press the button.  The Android app actually takes you to the introduction (if there is one) otherwise it takes you to chapter 1.  I personally wouldn't have an issue with just always going to chapter 1, but you definitely need to go to the beginning of the book in some fashion.

Exits from about, settings and layout are different than that of the android app.  If you exit those functions from Android, it takes you to the top level content menu when you return.  The PWA version is always returning to the Scripture view.  I don't have as much of an issue with that, but it is an inconsistency.

Finally, if a content menu exists, then it is always displayed at startup the first time the app starts.  If the launch option is set, on subsequent startups, it may show the contents menu or it may show the last book read, depending on the configuration.  But the first time, it always displays contents.  This isn't  a big deal, but it struck me because initially the Scripture view came up without an arrow until you pressed contents once.  The "hamburger" menu is never seen normally in the Scripture view if you have contents menu if you have the arrow configured.   Even if you restart and come up with the last read book displayed, you still see an arrow at the top of the screen which takes you to the initial contents menu.

src/routes/+page.svelte
```diff
@@ -304,11 +307,20 @@
...
+        if ($contentsStack.length > 0) {
+            const menuId = contentsStack.popItem();
+            goto(`${base}/contents/${menuId}`);
+        }
+    }
+    $: showBackButton = contents?.features?.['navigation-type'] === 'up' && $contentsStack.length > 0
```
davidmoore1, line 318: I suspect this is where the change would need to be made to make sure the back button always shows if the navigation type is up.  Then changes made so that if it is pressed and the stack is empty, it goes to the first contents screen.

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,205 @@
...
+    } from '$lib/data/stores';
+    import { base } from '$app/paths';
+    import { refs } from '$lib/data/stores';
+    import { goto } from '$app/navigation';
+    import config from '$lib/data/config';
+
+    $: highlightColor = $themeColors['ContentsItemTouchColor'];
+    $: title = setTitle($page);
+
+    console.log(config.interfaceLanguages);
```
Comment (davidmoore1) [inline src/routes/contents/[id]/+page.svelte:22]: Comment needs to be removed for final commit

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,205 @@
...
+
+    function setReference(item) {
+        let book;
+        let chapter;
+        let verse;
+        const reference = item.linkTarget.split('.');
+
+        switch (reference.length) {
+            case 1:
+                book = reference[0];
```
Comment (davidmoore1) [inline src/routes/contents/[id]/+page.svelte:86]: If book only is set, then the chapter needs to be set to 1 or introduction

src/routes/contents/[id]/+page.svelte
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    function handleBackNavigation(event) {
+        event.preventDefault();
+        if ($contentsStack.length > 0) {
+            const menuId = contentsStack.popItem();
+            goto(`${base}/contents/${menuId}`);
+        }
+    }
+    $: showBackButton = $contentsStack.length > 0;
```
Comment (davidmoore1) [inline src/routes/contents/[id]/+page.svelte:135]: This one is correct.  In the contents menu itself, the back arrow only shows if the stack is greater than zero.

davidmoore1 (changes requested):

chrisvire:
@mikaylahubbard: Let's fix the navigation and launch with the contents menu with a different issue.

Please look at the issue with navigation to Matthew (beginning of the book).

chrisvire:
@mikaylahubbard I cleaned the few remaining items.

chrisvire (approved):
Thank you for all your hard work!

davidmoore1 (approved):
Looks good

PR #570, opened 2024-05-24
Collection footer does not take on book footer

### LINKED ISSUE #569: Convert footer adds extra footer line to config.js
If you define a book collection without a footer, and you define a book within that collection as having a footer, SAB puts a single footer line in the configuration for the book and no footer entry for the book collection.  The config.js file that results after that is processed contains an entry for the book, but now also has the same entry for the book collection level.   The book collection entry should not be present.

gmjgeek:
Fixes #569

Previously, the conversion process would sometimes assign a book-specific footer to the entire collection. This generally happened when the collection-level footer was undefined.

Now, the conversion process correctly assigns footers to book collections as specified by the app.xml exported from SAB.

### DIFF
--- a/convert/convertConfig.ts
+++ b/convert/convertConfig.ts
@@ -372,6 +372,14 @@ function convertFooter(markdown: string | undefined, appdef: Document): string |
         ?.replace(/%program-version%/g, programVersion ?? '');
 }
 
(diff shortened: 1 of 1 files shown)


davidmoore1 (approved):
Looks good.  Thanks

PR #571, opened 2024-05-24
Fix quiz conversion process regarding collections

FyreByrd:
Per @chrisvire 's request, switched quiz conversion to store quizzes in a folder named after the parent collection instead of the docSet.

The primary rationale is that docSet is a Proskomma specific entity, rather than one specific to an SAB project, unlike book collections.

### DIFF
--- a/convert/convertBooks.ts
+++ b/convert/convertBooks.ts
@@ -227,7 +227,43 @@ export async function convertBooks(
         const frozen = freeze(pk);
         freezer.set(context.docSet, frozen[context.docSet]);
         //start catalog generation process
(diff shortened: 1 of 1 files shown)


FyreByrd:
Added some changes to address some issues arising from the edge case where a docSet has no documents.
This happens when a book collections does not have any books that would be processed by Proskomma.

chrisvire:
Let's have two separate PRs:
1) change the quiz path to be based on collection.id instead of docSet
2) handle empty docSet.

I believe that I had code in SAB that would keep from building PWA without scripture books.  It might be that I just checked for number of books regardless of book type (i.e. a bug).

chrisvire (approved):
👍 

Well done!

PR #580, opened 2024-05-28
Fix/empty docset

FyreByrd:
Per @chrisvire 's request, moved some changes from PR #571 to a new PR.

In some cases when attempting to implement quizzes, a data set was used which contained book collections that only had quizzes. This PR is an attempt to fix the problems that arose during the conversion process on those data sets.

While this fixes those problems, I am very certain that the chosen solution, not writing a pkf when there is no Proskomma docSet, will break many things in the app itself.

Pretty much everything in `src/lib/data/scripture.js` is likely to break from this change, which will also probably cause problems in `src/routes/+page.svelte`.

One potential solution that could fix those issues would be extending `src/lib/sab-proskomma` to be able to handle an empty pkf, and then make sure an empty pkf is written instead of no pkf.

The main reason this PR exists is because we have data sets that cause this problem. It may be the case that something should be changed in Scripture App Builder itself to prevent the creation of data sets that cause this problem in the first place.

### DIFF
--- a/convert/convertBooks.ts
+++ b/convert/convertBooks.ts
@@ -227,7 +227,43 @@ export async function convertBooks(
         const frozen = freeze(pk);
         freezer.set(context.docSet, frozen[context.docSet]);
         //start catalog generation process
-        catalogEntries.push(pk.gqlQuery(queries.catalogQuery({ cv: true })));
+        catalogEntries.push(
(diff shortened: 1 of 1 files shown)


FyreByrd:
This is a very old PR. I don't think this needs to stay open at this point.

Issue #605, opened 2024-06-17
Audio fails if audio source url ends in a slash character


davidmoore1:

benpaj-cps:
Closed by #606.

PR #606, opened 2024-06-17
Issue #605 - Audio fails if audio source is a url with a slash as the…

davidmoore1:
… last character.

### DIFF
--- a/src/lib/data/audio.ts
+++ b/src/lib/data/audio.ts
@@ -383,9 +383,13 @@ export async function getAudioSourceInfo(item: {
     } else if (audioSource.type === 'assets') {
         audioPath = '/audio/' + audio.filename;
     } else if (audioSource.type === 'download') {
-        audioPath = audioSource.address + '/' + audio.filename;
+        let extraSlash = '/';
+        if (audioSource.address.endsWith('/')) {
+            extraSlash = '';
+        }
+        audioPath = audioSource.address + extraSlash + audio.filename;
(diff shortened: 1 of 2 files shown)


src/lib/data/audio.ts
```diff
@@ -383,9 +383,13 @@ export async function getAudioSourceInfo(item: {
     } else if (audioSource.type === 'assets') {
         audioPath = '/audio/' + audio.filename;
     } else if (audioSource.type === 'download') {
-        audioPath = audioSource.address + '/' + audio.filename;
+        let extraSlash = '/';
```
chrisvire, line 386: Please make a more robust joinPaths function in stringUtils.ts?

  Look at:
https://stackoverflow.com/questions/29855098/is-there-a-built-in-javascript-function-similar-to-os-path-join

src/lib/data/audio.ts
```diff
@@ -383,9 +383,13 @@ export async function getAudioSourceInfo(item: {
...
+        if (audioSource.address.endsWith('/')) {
+            extraSlash = '';
+        }
+        audioPath = audioSource.address + extraSlash + audio.filename;
+        console.log('Audio Path:', audioPath, audioSource.address, audio.filename);
```
chrisvire, line 391: Remove console.log

chrisvire (changes requested):
Make a utility function?
Remove console.log?

chrisvire (approved):
Noice!

Issue #522, opened 2024-05-10
Edit Note popup should be full screen and include delete button.


davidmoore1:
The popup that is currently being displayed when the note is edited (either from pressing "Edit" in the notes window or the pencil icon on the note popup in the text) should be replaced by a full screen modal window similar to the native app.  It should have a back arrow in place of the "Cancel", a checkmark in place of "OK", and a trash can icon that will delete the note if pressed.  The existing note display popup with the pencil in the main text is good and should not be changed.

![Edit Popup](https://github.com/sillsdev/appbuilder-pwa/assets/6209188/1e9842ea-2478-45a3-a55d-51fb553bcca1)

PR #610, opened 2024-06-21
Feature/Edit Note make Fullscreen and Delete Button #522

Mast3Rei:
… and DeleteIcon for saving and deleting notes; added a route in /notes/+page.svelte to the edit route; attempted to redirect the in-text note editor pop-up to the fullscreen editor when the pencil icon is clicked


src/lib/components/NoteDialog.svelte
```diff
@@ -3,80 +3,55 @@
...
     let id = 'note';
     let modal;
-    let title: string;
     let text: string;
 
-    $: heading = editing ? $t[title] ?? '' : note?.reference ?? '';
+    $: heading = note?.reference ?? '';
 
-    export function showModal() {
+    export async function showModal() {
```
chrisvire, line 18: There is no await, so it doesn't need to be async.

chrisvire:
The title is small?

image

chrisvire:
What size text should the new note textarea use?  Is it the body size?

chrisvire (changes requested):
Need to deal with font sizes.

Issue #603, opened 2024-06-13
Video options


chrisvire:
The feature `video-youtube-related-same-channel` is changed in the Media > Video > Video Settings page:

![Image](https://github.com/sillsdev/appbuilder-pwa/assets/517814/dd3d46c2-f3eb-41b4-83c2-4024e0de8fac)

It looks like we are not accounting for these settings (in `config.mainFeatures`) as well:

```
    "video-allow-fullscreen": true,
    "video-show-titles": true,
    "video-full-screen-width": true,
```

davidmoore1:
The "Use the full width of the screen" option sets feature "video-full-screen-width".  In the Android app, this causes it to override the margin-left and margin-right which are used for the div.video-block style with the values that are set for the body style.  In our case, this would be the styles for the #container.  So find the left and right margins associated with the container, and set those values as the left and right margins for the video div.

For the "video all full screen" option "video-allow-fullscreen", investigate whether it is possible to do this.  For the android app, he only uses this when using a chrome client view.  There is a custom chrome web view that derives from the WebChromeClient.  There is an override of onShowCustomView and onHideCustomView that is called when the user presses the full screen button from within the embedded video.  In that method, it looks like he creates a view of the same size as the original if they are not allowed to use full screen.  I don't know if it is possible to catch the embedded video view button press within PWA, particularly when you have no control of the type of browser being used.

PR #611, opened 2024-06-24
Configure Related Videos and Title Under Videos

Christopher-Shiu:
Issue #603
Problem: Full Screen and margins not done

### DIFF
--- a/src/lib/video/index.ts
+++ b/src/lib/video/index.ts
@@ -52,7 +52,6 @@ export function createVideoBlock(document: Document, video: any, index: any): HT
             videoContainerDiv.classList.add('video-YouTube', 'video-16-9');
             break;
         case VideoType.Mp4:
-        case VideoType.YouTube:
         case VideoType.Vimeo:
         case VideoType.Hls:
         case VideoType.Gospel:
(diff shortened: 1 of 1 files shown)


davidmoore1 (approved):

src/lib/video/index.ts
```diff
@@ -39,6 +41,12 @@ export function createVideoBlock(document: Document, video: any, index: any): HT
     const id = 'VIDEO' + index;
     videoContainerDiv.setAttribute('id', id);
     videoContainerDiv.classList.add('video-container');
+
+    let videoUrl = video.onlineUrl;
+    if (type === VideoType.YouTube && !config.mainFeatures['video-youtube-related-same-channel']) {
```
davidmoore1, line 46: you know, I think I would change and incorporate it with the switch that is just to follow.  Make case VideoType.YouTube:
if (!config.mainFeatures['video-youtube-related-same-channel'] {
videoUrl +=  etc
}
videoContainerDiv.classList.add ....
break;
case VideoType.mp4:
etc

davidmoore1 (changes requested):

davidmoore1 (approved):

Issue #582, opened 2024-05-30
Contents menu can have audio clips


davidmoore1:
Prerequisite: Issue #536 Basic contents menu

Audio clips may be configured to play for specific contents menu items.  Each item that has an audio associated with it displays a small speaker icon as part of the contents display:

![ContentsWithAudio](https://github.com/sillsdev/appbuilder-pwa/assets/6209188/8ed01cb8-3675-40a8-9058-bc045ed03d64)

The audio plays when the user presses the speaker button.  If it is not pressed, the audio is not played.

There may be a single audio clip for the entry, or more than one for different interface languages.  The audio clips are specified in the contents.xml file.

An example for an entry with a single audio file defined: 
```xml
    <contents-item id="1">
      <title lang="default">Noah and the Great Flood</title>
      <audio lang="default">Noah.mp3</audio>
      <image-filename>01_Ge_07_08.png</image-filename>
      <link type="reference" target="GEN.6" />
      <layout mode="single">
        <layout-collection id="C01"/>
      </layout>
    </contents-item>
```

An example for an entry with multiple audio files defined:
```xml
<contents-item id="3">
      <title lang="default">Jesus Parables</title>
      <title lang="es">Las parábolas de Jesús</title>
      <audio lang="en">parables_en.mp3</audio>
      <audio lang="es">parables_es.mp3</audio>
      <image-filename>parables.png</image-filename>
      <link type="screen" target="2" />
    </contents-item>
```

If no entry is defined for the current language (like russian is a defined interface language for this app, but no audio clip for that has been provided), the first audio (English) plays when the button is pressed when Russian is the interface language.

PR #619, opened 2024-07-01
play contents menu audio clips

mikaylahubbard:
- fixes issue #582
- adds the audio icon in the appropriate place if there is audio in the contents menu item
- changes the onclick to go to the playAudio() fuction instead of navigating if the
icon is clicked
- playAudio() function plays the correct audio based on the interface language
- also changes the formatting of the html to use <div> instead of <a> as per an update in the css
- potential issue: the way I go about playing the audio instead of navigating could be too
complicated/might not be neccessary if there is a
different way it should work (z-index?)

### DIFF
--- a/src/routes/contents/[id]/+page.svelte
+++ b/src/routes/contents/[id]/+page.svelte
@@ -27,19 +27,12 @@
     function playAudio(event, item) {
         event.stopPropagation();
         //assign/use the proper filename
-        let filename;
-        if (item.audioFilename[$language]) {
-            filename = item.audioFilename[$language];
-        } else {
-            filename = item.audioFilename.default;
-        }
+        const filename = item.audioFilename[$language] ?? item.audioFilename.default;
 
         let audio = new Audio();
         audio.src = `${base}/${audioFolder}/${filename}`;
         audio.play();
     }
-
-    //console.log(config.interfaceLanguages);
 
     function onClick(event, item) {
         event.target.style.background = highlightColor;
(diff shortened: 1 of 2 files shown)


src/routes/contents/[id]/+page.svelte
```diff
@@ -16,16 +16,35 @@
...
     const imageFolder =
         compareVersions(config.programVersion, '12.0') < 0 ? 'illustrations' : 'contents';
     const audioFolder = compareVersions(config.programVersion, '12.0') < 0 ? 'assets' : 'contents';
     $: highlightColor = $themeColors['ContentsItemTouchColor'];
     $: title = setTitle($page);
 
+    function playAudio(event, item) {
+        event.stopPropagation();
+        //assign/use the proper filename
+        let filename;
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:30]: You can use a null coalescing operator to make this more concise.

```
const filename = item.audioFilename[$language] ?? item.audioFilename.default;
```

src/routes/contents/[id]/+page.svelte
```diff
@@ -16,16 +16,35 @@
...
+        } else {
+            filename = item.audioFilename.default;
+        }
+
+        let audio = new Audio();
+        audio.src = `${base}/${audioFolder}/${filename}`;
+        audio.play();
+    }
+
+    //console.log(config.interfaceLanguages);
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:42]: Remove the console.log

src/routes/contents/[id]/+page.svelte
```diff
@@ -152,58 +171,67 @@
...
+                                item
+                            )}"
+                        >
+                            <img
+                                class="contents-image"
+                                src="{base}/{imageFolder}/{item.imageFilename}"
+                                alt={item.imageFilename}
+                            />
+                        </div>
+                        <!-- Example of using item.audioFilename. Remove and replace with icon. -->
```
Comment (chrisvire) [inline src/routes/contents/[id]/+page.svelte:206]: Remove this example code that I left for you.

chrisvire (changes requested):
Good job! Just a few changes.

chrisvire (approved):
Looks good! Thanks!

PR #636, opened 2024-07-15
Search - Return with Back Button

gmjgeek:
Allows the user to return to the latest search results using the back button.

Features:
* Uses a query parameter in the URL to tell when the user is returning through the back button
* When the user returns through the back button, the page loads results from the previous search and scrolls to the last saved position
* To conserve storage, searches are limited to 500 results
* Works with Android back button and browser back button

Search results are saved using localStorage and cleared with every new search.


chrisvire (changes requested):
This is good. I think that llimiting to 500 is not necessary.  Maybe 1000?

chrisvire (approved):
Well done!

PR #637, opened 2024-07-15
Define styles for compatability in Slider.svelte

gmjgeek:
Fixes svelte-check warnings about CSS compatibility in Slider.svelte

Defines the `appearance` style to match `-webkit-appearance`


chrisvire (approved):
Well that was easy. I wonder why I didn't read that warning a little more closely... sigh.

Issue #537, opened 2024-05-14
Contents menu text size should be adjustable


davidmoore1:
Prerequisite: Issue #536 

If the text appearance screen is selected while displaying the contents menu, it displays with two options, the text size option and the color theme selection bar.  The font selection is not available when displaying contents.  The font size is the font size of the contents title.   In the app, the size is coming from the body.contents css style.  As the size is adjusted up and down, the size of the contents menu text should change.  The reference and subtitles are defined as a percentage of the main body contents size.  This change should only affect the content screen, not the size of the Bible text.  And the reverse is also true, changing the size of the Bible text should not affect the content text size.

Theme change can be done from this menu as well, changing to any of the three styles.  These changes do apply to all screens in the app, not just the content screens.  Likewise the style changes done while in the Bible text also change the theme in the content window.

![Content-TextSize](https://github.com/sillsdev/appbuilder-pwa/assets/6209188/749762fd-e7a5-449a-822a-a0e023cca705)

PR #646, opened 2024-07-19
add text appearance selector to contents menu

mikaylahubbard:
fixes issue #537, contents text size should be
adjustable
- shows the modal for text TextAppearance if the setting is true
- adds a contentsMdoe to TextAppearanceSelector
- shows only the options that need to be shown if its in contents mode
- creates a seperate contentsTextSize


chrisvire (approved):
Looks good!

Issue #632, opened 2024-07-12
Do not allow swiping between books if option is set


davidmoore1:
Currently the < > buttons will always transition to the next or previous book if you are at the end or beginning of a book.  If the option to allow swipe between books, set in Book Navigation section of SAB, is not checked, then the < or > button should not appear and you should not be able to swipe between books.  The option is controlled by "book-swipe-between-books" in main-features.  If it is false, you should not be able to swipe between books.

PR #647, opened 2024-07-19
Optionfix/disable swipe between books #632

Mast3Rei:
In the src/routes/+page.svelte modifications are made to the doSwipe function to read in the 'Allow Swiping Between Books' feature and prevent swiping between books if the option is disabled

### DIFF
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -72,48 +72,16 @@
         savedScrollPosition = 0;
     });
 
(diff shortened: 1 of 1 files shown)


davidmoore1 (changes requested):
This functionally still has some issues I think.  It does prevent swiping between books.  However the "arrow" keys on the side do not seem to be affected and I could still press them to go from book to book.  They should also be responsive to the same option and that was mentioned in the original issue.

PR #653, opened 2024-07-23
Support Contents Menu Launch Action

chrisvire:
* move scripture display to /text route
* fix firstLaunch so that it can be handled by start page and text page
* handle query string parameters in start page
* use navigate function to add query string ref to history


src/routes/text/+page.svelte
```diff
@@ -0,0 +1,479 @@
...
+    afterUpdate(() => {
+        if (scrollingDiv) {
+            scrollingDiv.scrollTop = savedScrollPosition;
+        }
+    });
+    refs.subscribe((value) => {
+        savedScrollPosition = 0;
+    });
+    async function doSwipe(event) {
+        console.log('SWIPE', event.detail.direction);
```
davidmoore1, line 74: Do you want to leave this in?

davidmoore1 (approved):
Looks fine with the one comment about the console.log line that may not be wanted.

PR #655, opened 2024-07-29
Fix deprecated call to filesystem

gmjgeek:
Replaces call to fs.rmdirSync() with fs.rmSync().

See npm deprecation [DEP0147](https://nodejs.org/api/deprecations.html#DEP0147)


chrisvire (approved):
👍

PR #656, opened 2024-08-02
Disable swipe between books when config is diabled #632

Mast3Rei:
In the src/routes/text/+page.svelte modifications are made to read in the 'Allow Swiping Between Books' feature and prevent swiping between books and make the proper arrows invisible if the option is disabled


chrisvire (approved):
Good job!  Thanks!

PR #659, opened 2024-08-06
Feature/quiz score history 2

aidanpscott:


src/routes/quiz/[collection]/[id]/+page.js
```diff
@@ -1,21 +1,58 @@
...
+        }
     }
+
+    return {
+        locked,
+        quizId: id,
+        quizName: book.name,
+        dependentQuizId,
+        dependentQuizName,
+        passScore: 0
```
Comment (chrisvire) [inline src/routes/quiz/[collection]/[id]/+page.js:56]: Why returning "passScore: 0" ?

Comment (chrisvire) [reply src/routes/quiz/[collection]/[id]/+page.js:56]: If it is locked then the passScore doesn't matter?  Why is there two separate returns?

Comment (chrisvire) [reply src/routes/quiz/[collection]/[id]/+page.js]: I have fixed to have a single return statement

chrisvire (approved):
👍

PR #666, opened 2024-08-13
Use correct allow-single-pane feature in 12.1

chrisvire:
In 12.1, the feature name that controls whether a collection is allowed in single-pane mode changed
from: bc-allow-single-pane
to:   bc-layout-allow-single-pane

### DIFF
--- a/convert/convertConfig.ts
+++ b/convert/convertConfig.ts
@@ -1235,9 +1235,12 @@ function filterFeaturesNotReady(data: ConfigData) {
     data.mainFeatures['settings-share-usage-data'] = false;
 
     if (data.bookCollections) {
-        //only allow single pane book collections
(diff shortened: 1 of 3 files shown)


davidmoore1 (approved):
I think this works although it doesn't completely match the apps.  The bc-allow-single-pane actually changed to bc-layout-restricted and flipped from being true if allowed to be false if allowed.  The bc-layout-allow-single-pane is a new option unconnected (technically) to restricted which says whether the book collection is allowed to be shown in single mode.  The app checks that that is true and the other is false to allow it to be displayed.  In fact, SAB appears to be setting the bc-layout-allow-single-pane to false whenever you set the restricted flag to true, so you probably don't have to check both flags.

PR #695, opened 2024-09-18
Issue #694 - Footnote attached to /s crashes page display

davidmoore1:
Redid the way the text was being handled so that the footnote could be properly added to the header information


davidmoore1:
Issue #694

Issue #681, opened 2024-09-11
Convert: run different scripts based on project type


chrisvire:
SAB and RAB have different sets of files that need to be converted.  Modify the `convert/index.ts` script to convertConfig first and then decide which conversions to run based on the project type.

PR #702, opened 2024-10-24
#681 Run Convert Scripts Based on Project Type

EthanFennell:
Some convert scripts are only necessary for SAB, and some in the future will only be neccessary for DAB. This pull request:

-Separates the list of convert scripts into common, SAB only, and a future DAB only
-Uses programType from ConvertConfig (which comes from appdef.xml) to determine which lists to use


chrisvire (approved):
👍

PR #713, opened 2024-11-01
Display html books

chrisvire:
- [x] Convert HTML books
  - [x] Update image tags to reference `/illustrations`
  - [x] Use BUILD_BASE_PATH for path to illustrations
- [x] HTML books display in the Book Selector 
- [x] When an HTML book is selected
  - [x] Display like scripture (with gutters on desktop)
  - [x] Chapter selector hidden
  - [x] Display title in book selector
  - [x] No swiping or next/prev buttons
  - [x] Use styles in `<head>` of the HTML file
  - [x] Display correctly on mobile
  - [x] Font size adjustment

Sample project provided by user: https://www.evangelischebijbelvertaling.nl/EBV.zip

### DIFF
--- a/convert/convertBooks.ts
+++ b/convert/convertBooks.ts
@@ -12,6 +12,8 @@ import { convertMarkdownsToMilestones } from './convertMarkdown';
 import { verifyGlossaryEntries } from './verifyGlossaryEntries';
 import { hasAudioExtension, hasImageExtension } from './stringUtils';
 
+const base = process.env.BUILD_BASE_PATH || '';
+
 /**
  * Loops through bookCollections property of configData.
  * Each collection and all associated books are imported into a SABProskomma instance.
@@ -114,6 +116,18 @@ function removeMissingFigures(text: string, _bcId: string, _bookId: string): str
     });
 }
 
+function updateImgTags(text: string, _bcId: string, _bookId: string): string {
+    return text.replace(
+        /<img\b[^>]*\bsrc=["']([^"']*\/)?([^"']*)["'][^>]*>/gi,
+        (_match, _path, fileName) => {
+            const imagePath = `${base}/illustrations/${fileName}`;
+
+            // If the image is missing in the "illustrations" folder, filter out the entire tag
+            return isImageMissing(fileName) ? '' : `<img src="${imagePath}">`;
+        }
+    );
+}
+
 function trimTrailingWhitespace(text: string, _bcId: string, _bookId: string): string {
     return text
         .split('\n') // Split the text into lines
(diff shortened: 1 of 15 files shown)


chrisvire:
I need to fix display on moble!

image

Here it is on desktop:

image

src/lib/data/navigation.ts
```diff
@@ -119,8 +122,17 @@ export class NavigationContext {
 
     private updateHeadings() {
         const document = this.catalog.documents.find((b) => b.bookCode === this.book);
-        this.title = document.toc;
-        this.name = document.h;
+        if (document) {
+            this.title = document.toc;
+            this.name = document.h;
+            return;
```
davidmoore1, line 128: Are you still not going to find the html book when searching the catalog?  Not sure what this and the next one are doing

src/lib/data/stores/scripture.js
```diff
@@ -151,11 +151,11 @@ export const fontChoices = derived(refs, ($refs) => {
     if (!$refs.initialized) return [];
     const bookFonts = config.bookCollections
         .find((x) => x.id === $refs.collection)
-        .books.find((x) => x.id === $refs.book).fonts;
+        .books.find((x) => x.id === $refs.book)?.fonts;
     const colFonts = config.bookCollections.find((x) => x.id === $refs.collection).fonts;
     const allFonts = [...new Set(config.fonts.map((x) => x.family))];
     const currentFonts =
-        bookFonts.length > 0 ? bookFonts : colFonts.length > 0 ? colFonts : allFonts;
+        bookFonts?.length > 0 ? bookFonts : colFonts.length > 0 ? colFonts : allFonts;
```
davidmoore1, line 158: what happens to this comparison if there is no book fonts.  Is it undefined?

src/lib/scripts/configUtils.ts
```diff
@@ -64,6 +64,7 @@ export function getStyle(configData: any, option: string, bc: string, book: stri
     const bookData = bcData.books.find((x) => x.id === book);
     // Use the style of the book, if defined.
     if (
+        bookData &&
```
davidmoore1, line 67: I had just hit this one too yesterday.  Looks good.

davidmoore1 (approved):
Generally, approve with questions

convert/convertBooks.ts
```diff
@@ -386,6 +420,20 @@ export type Quiz = {
...
+    const dstFile = path.join('static', 'collections', context.bcId, book.file);
+
+    let content = fs.readFileSync(srcFile, 'utf-8');
+    content = applyFilters(content, htmlFilterFunctions, context.bcId, book.id);
+    files.push({
+        path: dstFile,
+        content
+    });
+
+    //process.stdout.write(`copyHtmlBook: bookId:${book.id}, src:${srcFile}, dst:${dstFile}\n`);
```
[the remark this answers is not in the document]
chrisvire: Can be deleted.

src/lib/data/navigation.ts
chrisvire: The old code was pulling out the `title` and `name` properties from the document in the catalog. I tried to keep it consistent.  However, I don't see where these are even being used.

davidmoore1 (approved):

PR #715, opened 2024-11-07
Update README for installing convert dependencies

gmjgeek:
To fully update dependencies, one must run `npm install` in the root directory and in convert/.

Updated README.md to document this procedure


chrisvire:
@gmjgeek, @7dev7urandom :

 I did the following:

1. clone the repo to a new folder
2. just did `npm i` in the root folder
3. generated data to the `/data` folder with SAB
4. ran `npm run convert` 

It generated files just fine.

I tried to delete the `/convert/package*.json` files and `npm run convert` failed to run.
I tried to remove the dependencies in `/convert/package.json` and ran npm i to generate an "empty" `package-lock.json` file. I was able to run `npm run convert` from the root directory.

Do we need dependencies in `/convert/package.json` ?

@gmjgeek When was it failing and you needed to run `npm i` in the `/convert` directory ?

7dev7urandom:
You should not need to run `npm i` in the `/convert` directory. As you noticed, the `package.json` file is necessary anyway. Just make sure any dependencies needed by the convert scripts are listed in the root `package.json` (which they are) and everything will work fine.

chrisvire:
In https://github.com/sillsdev/appbuilder-pwa/pull/740, I removed the dependencies from the convert/package.json and removed the "npm i" in the convert directory (in the GitHub Action Workflows).

Issue #676, opened 2024-09-10
Convert: DAB: Parse Specific Config


chrisvire:
SAB and DAB share a lot of the same config, but there will be parts that are unique to DAB.

The main data elements in a DAB project are:

```
    <database-filename>data.sqlite</database-filename>
    <lexicon-db type="flex-xhtml" />
```

The main data elements in an SAB project are:

```
    <books id="C01">
        <book-collection-name>World English Bible</book-collection-name>
        <book-collection-abbrev>web</book-collection-abbrev>
        <book-collection-description></book-collection-description>
        ...
    </books>
```

You will need to make the `convert/convertConfig.ts` file work with both types.  The way that you tell which is which is in the `<app-definition>` tag:

```
<app-definition type="DAB" program-version="12.1">
```

The writing-system entries in DAB are more complex that what is in SAB.

Here is an example writing-systems from the Hanga project:
```
    <writing-systems>
        <writing-system code="hag-Latn-GH-fonipa-x-emic" type="main|gloss">
            <display-names>
                <form lang="default">Hanga</form>
            </display-names>
            <font-family>font-hag-Latn-GH-fonipa-x-emic</font-family>
            <trait name="text-direction" value="LTR"/>
            <trait name="indexed" value="false"/>
            <sort-method type="none">
                <ignore-chars>- = ~ * ( ) [ ] Â¡ Â¿</ignore-chars>
            </sort-method>
            <alphabet>a b ch d e f g gb h i j k kp l m n ny ŋ ŋm o p r s t u v w y z</alphabet>
            <input-buttons>a b ch d e f g gb h i j k kp l m n ny ŋ ŋm o p r s t u v w y z</input-buttons>
            <features type="writing-system">
                <e name="summary-gloss" value="true"/>
            </features>
        </writing-system>
        <writing-system code="en" type="gloss">
            <display-names>
                <form lang="default">English</form>
            </display-names>
            <font-family>font-en</font-family>
            <trait name="text-direction" value="LTR"/>
            <trait name="indexed" value="true"/>
            <sort-method type="none">
                <ignore-chars>- = ~ * ( ) [ ] Â¡ Â¿</ignore-chars>
            </sort-method>
            <alphabet>a b c d e f g h i j k l m n o p q r s t u v w x y z</alphabet>
            <reversal-filename>Hanga dictionary reversals.xhtml</reversal-filename>
        </writing-system>
    </writing-systems>
```

Here is an example writing-systems from SAB:
[...]

PR #725, opened 2024-11-18
Parsed Dictionary Attributes for writing systems

Christopher-Shiu:
Issue #676
Split parseInterfacelanguage into three functions
and modified index.d.ts as well as the Dictionary
Config function


convert/convertConfig.ts
```diff
@@ -727,6 +696,88 @@ function parseInterfaceLanguages(
...
+    if (alphabetTag) {
+        alphabet = alphabetTag.textContent?.split(/\s+/).filter((letter) => letter);
+    }
+
+    let inputButtons: string[] | undefined;
+    const inputButtonsTag = element.getElementsByTagName('input-buttons')[0];
+    if (inputButtonsTag) {
+        inputButtons = inputButtonsTag.textContent
+            ?.split(/\s+/)
+            .filter((button) => button);
```
chrisvire, line 748: Why is the `.filter` there?

chrisvire: These `.filter` calls are not needed.

PR #734, opened 2024-11-26
Run App Builders during Test Workflow

chrisvire:
Separate the testing from the Lint workflow. Run App Builders from the latest release against projects instead of depending on `example_data` folder, which is stale.

Include test projects in `test_data/projects` for SAB and DAB.

Add `npm run extract:example <project_name>` to use test projects for development.


chrisvire:
@davidmoore1 I would like to review this with you. Should we get rid of the example_data? Or how do we keep it up to date? The best would be to have a script in package.json to generate data from the /test_data/projects, but it would have to be cross-platform...

davidmoore1 (approved):
As we said during the meeting, it would be good to be able to come up with a replacement for the build examples that builds the checked in projects.  But that can be done as a separate PR.

chrisvire:
> But that can be done as a separate PR.

This was too much fun to wait for another PR.  Thanks you ChatGPT:
https://chatgpt.com/share/6747aa90-542c-8006-8d69-985b3a1768fb

PR #748, opened 2024-12-04
Remember font per book collection

chrisvire:


davidmoore1 (approved):
This works fine.  There is a separate issue that we aren't following the configuration for the ui that goes with the language.  I'll open a separate issue for that.

PR #769, opened 2025-01-12
Convert Navbar to Svelte 5

chrisvire:


src/routes/about/+page.svelte
```diff
@@ -13,11 +13,11 @@
 <div class="grid grid-rows-[auto,1fr]" style="height:100vh;height:100dvh;">
     <div class="navbar">
         <Navbar>
-            <!-- <div slot="left-buttons" /> -->
-            <label for="sidebar" slot="center">
-                <div class="btn btn-ghost normal-case text-xl">{$t['Menu_About']}</div>
-            </label>
-            <!-- <div slot="right-buttons" /> -->
+            {#snippet center()}
+                <label for="sidebar">
```
chrisvire, line 17: @davidmoore1: Here is an example of where there is a `<label for="sidebar">` so that if you tap/click on the about title, it shows the sidebar menu.

Positive: you can get to the menu without having to go back and then click on the hamburger menu.
Negative: It is different than the mobile apps. It is a hidden feature.

src/routes/text/+page.svelte
```diff
@@ -346,88 +345,89 @@
...
+                                    <AudioIcon.Volume color="white" />
+                                {:else}
+                                    <AudioIcon.Mute color="white" />
+                                {/if}
+                            </button>
+                        {/if}
+                    </div>
+                    <div id="extraButtons" class={showOverlowMenu ? 'flex' : 'hidden md:flex'}>
+                        <!-- An overflow menu containing the other right-buttons. On mobile it expands when overflowMenuButton is clicked and collpases when handleMenuClick() is called, on larger screens these buttons are always visible. -->
+
```
chrisvire, line 383: This is an incorrect diff when looking at it in GitHub

src/routes/about/+page.svelte
davidmoore1: I would say leave it in.  We don't have the swipe in capability to get the sidebar so this is a good alternative I think

davidmoore1 (approved):

PR #772, opened 2025-01-15
Hangs when selecting book with no chapter 1

davidmoore1:


chrisvire (approved):

chrisvire:
If there are other places that we are assuming that the first chapter is chapter 1, then we might want to add it to $refs object.

davidmoore1:
There are issues with at least one of the other two places.  I'll look creating a fix for those in a separate PR

PR #780, opened 2025-01-23
Allow for non string features in method signature.

davidmoore1:
The method signature called for an array of strings.  The features include boolean types and perhaps others.   Changing the type to any removes errors being detected in the locations where it is called.


chrisvire (approved):
👍

PR #783, opened 2025-01-23
Crash on reading plan book selected

davidmoore1:
This fixes a crash on trying to process a reading plan made of tables.  We still have an issue that for this case, selecting chapter 1 is still returning the entire book from Proskomma


chrisvire (approved):
Better than before. I think the receiving all chapters is a separate issue that I will bring up today with Mark.

Issue #781, opened 2025-01-23
Content Menu References are not displayed correctly


davidmoore1:
They are currently displayed as the reference abbreviations in the contents.xml file.  Those should be resolved to text.

benpaj-cps:
Closed by #782.

PR #803, opened 2025-04-04
Issue #781 Fix problem with contents reference causing contents to not display

davidmoore1:
Previous change broke contents menu because async function wasn't handled properly

## Summary by CodeRabbit

- **New Features**
  - Enhanced reference text loading with an asynchronous caching mechanism, ensuring texts are fetched and displayed smoothly. The UI now handles dynamic loading states, providing a more efficient and responsive experience.


chrisvire (approved):
👍

chrisvire (changes requested):
Please use `{:then reference}` to capture the output and have loadReference return the current value.

PR #806, opened 2025-04-25
Feature/search dictionary

AslanRules738:
## Summary by CodeRabbit

- **New Features**
  - Introduced a dedicated lexicon search page with advanced search options, including accent and whole word matching.
  - Added batch display of lexicon entries and improved navigation between search results.
  - Enhanced clickable word links and styling in lexicon entry views.
- **Improvements**
  - Unified language and word list management using centralized language stores for a more consistent experience.
  - Updated sidebar navigation to handle program-specific search routing.
  - Improved layout and scroll handling for lexicon pages.
  - Applied consistent background and text color styling to list, alphabet, and navigation strip views using CSS variables.
  - Refined search form styling and conditional display based on program type.
  - Enhanced navigation strip styling and internal data sourcing from stores.
- **Bug Fixes**
  - Ensured correct handling of accent and tone search options.
- **Chores**
  - Updated dependency for "sql.js" to a newer version.


src/lib/components/LexiconEntryView.svelte
```diff
@@ -1,67 +1,49 @@
...
-            if (stmt.step()) {
-                result = stmt.getAsObject().xml;
-            }
-            stmt.free();
-            db.close();
+            let results;
+            const dynamicQuery = wordIds.map(() => `id = ?`).join(' OR ');
+            const dynamicParams = wordIds.map((id) => id);
+            results = db.exec(`SELECT xml FROM entries WHERE ${dynamicQuery}`, dynamicParams);
+            console.log('results:', results[0].values);
```
chrisvire, line 26: Remove debug console.log

chrisvire:
Here is some UX feed back:
* I don't like the highlighting of the alphabet characters in the LexiconAlphabetStrip when you tap on a letter. It is visually distracting. Also, it is wrong at times.
* It might be slowing down the update of the page to change the position of the scrolling list. It seems a lot slower.
* The search result are coming back in a different order (e.g. search for "kpaama")
* The xml doesn't look correct

I will merge this PR and create issues for these items.

Issue #804, opened 2025-04-15
Support Multiple Reversals


chrisvire:
A Dictionary app can have multiple reversals. I found a project that uses this feature. Akoose Dictionary.
https://drive.google.com/file/d/12dOY8vtkx_w8NFVBU_BgaqaJruuHYleM/view?usp=drive_link

PR #807, opened 2025-05-01
Multiple Reversals completed implementation

Christopher-Shiu:
Issue #804
Transferred code from different branch
and fixed small errors with letter loading


chrisvire (changes requested):
Good job with getting the languages to display. I see that it changes from the vernacular language to the another one of the reversal languages, but switch between reversal languages just changes the letters at the top (and I am not sure that they are correct).

I am using this project:
https://drive.google.com/file/d/1YL65KBat-ioMgQYm4u5jPzTcIWUCZ1hQ/view?usp=sharing

PR #818, opened 2025-05-27
Fix issues encountered trying to setup the dev env

hahn-kev:
The main issue I had was that the format of the data in `index.json` had drifted from the type that was being used in the script. Instead of just correcting it I imported the file directly. 

Additionally I don't have `java` on my path, but I do have a standard `JAVA_HOME` env variable, so I updated the script to use that to find the java exe if it exists.


example/index.ts
```diff
@@ -30,7 +24,11 @@ const getExecutionCommand = (program: string): string => {
                 throw new Error('Environment variable "ProgramFiles(x86)" is not set on Windows');
             }
             const jarPath = path.join(programFilesX86, 'SIL', appName, 'bin', jarName);
-            return `java -jar "${jarPath}"`;
+            let javaPath = 'java';
```
chrisvire, line 27: This only deals with the case with java not being on the path for Windows.

hahn-kev: I can move it out so it works anywhere.

README.md
```diff
@@ -19,7 +19,7 @@ The PWA depends on data files generated by Scripture App Builder. There is examp
 
 #### Example Data
 
--   Run `npm run extract:example <project_name>` to generate data from projects in the `test_data/projects`. For example, `web_gospels` is an SAB project name and `hanga` is a DAB project name.
+-   Run `npm run extract:example <project_name>` to generate data from projects in the `test_data/projects`. For example, `web_gospels` is an SAB project name and `hanga` is a DAB project name. Note, this requires java to either be on the path, or to have a JAVA_HOME env variable on windows
```
chrisvire, line 22: We try not to do things that only work on one platform, so if we implement this, we would remove the comment about windows as well.

example/index.ts
```diff
@@ -4,15 +4,9 @@ import fs from 'fs/promises';
 import os from 'os';
 import path from 'path';
 import unzipper from 'unzipper';
-
-type IndexData = {
-    [key: string]: {
-        projects: string[];
-    };
-};
+import IndexData from '../test_data/projects/index.json';
```
chrisvire, line 7: I would prefer to change the datatype than do the import of a .json file.

chrisvire: I imagine an AI could easy discern the required change in datatype. :-)

hahn-kev: Is there a reason you don't want to infer the type from the file? We could do `import type IndexData from '../test_data/projects/index.json';` instead here and then read the file in from the file system at runtime like we were before.

chrisvire (approved):
Thanks for finding this!

Issue #819, opened 2025-06-04
Svelte 5 Migration


chrisvire:
Follow https://svelte.dev/docs/svelte/v5-migration-guide to migrate code from v4 to v5

Iniital upgrade to Svelte 5, but still using Svelte 4 Syntax:
- #766 

Previous migration work:
- #769  

Unmerged attempt at auto migration
- #762

judah-sotomayor:
# List of Files that Need Updating
- [ ] Test?

TheNonPirate:
Currently in-progress by me but not made into a PR yet:
- BookSelector.svelte
- ChapterSelector
- SelectList
- SelectGrid
- TabsMenu

judah-sotomayor:
> In Progress:
> 
> * FontSelector
> * FontList
> * HistoryCard

Any movement on these?

joslane:
Migrated the entirety of src/routes across these PRs: #884 #886 #887 #888 #889

chrisvire:
Stores are not being removed.

FyreByrd:
> Stores are not being removed.

Noted. Mostly just trying to think through how to find remaining files...

judah-sotomayor:
Over winter break I can look at some solutions for this, some of the linters may be able to warn us about deprecated syntax

FyreByrd:
The following commands should be helpful in finding files to update:
`git grep "\$:"` (replaced by $derived and $effect)
`git grep "<slot"` (replaced by snippets)
`git grep "export let"` (replaced by $props)
`git grep "\son:"` (legacy event handler syntax)

PR #833, opened 2025-06-10
Migrate horizontal pane and HTML page to Svelte 5

joslane:
Update horizontal pane by utilizing props for states. Implement $derived() runes. Change pointer functions to remove colon.

Update HTML page replacing reactive statements with $effect() and $derived() runes. Change exports to props.


joslane:
Continues Issue #819

PR #835, opened 2025-06-11
Update ScriptureViewSofria to Svelte 5

TheNonPirate:
Per #819 

This PR updates ScriptureViewSofria to Svelte 5.

There is a warning for the variables scriptureLogs and bookRoot saying, "This reference only captures the initial value of [the variable]. Did you mean to reference it inside a closure instead?"
As far as I can tell, this is not an actual problem, but I'm still getting used to how reactivity in Svelte works, so there might be a legitimate problem here.

## Summary by CodeRabbit

- **Refactor**
  - Improved component reactivity and state management for enhanced performance and clarity, with no changes to user-facing features or behavior.

### DIFF
--- a/src/lib/components/ScriptureViewSofria.svelte
+++ b/src/lib/components/ScriptureViewSofria.svelte
@@ -182,12 +182,6 @@ LOGGING:
         }
     });
 
-    $effect(() => {
-        if (references) {
-            observeVisibility();
-        }
-    });
-
     onDestroy(() => {
         if (planDivObserver) {
             planDivObserver.disconnect();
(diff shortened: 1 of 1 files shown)


src/lib/components/ScriptureViewSofria.svelte
```diff
@@ -156,7 +182,7 @@ LOGGING:
         }
     });
 
-    afterUpdate(() => {
```
chrisvire, line 159: How did you determine that the correct conversion of `afterUpdate` is to use `$effect`?

TheNonPirate: According to this link: https://svelte.dev/docs/svelte/lifecycle-hooks
"Instead of beforeUpdate use $effect.pre and instead of afterUpdate use $effect instead - these runes offer more granular control and only react to the changes you’re actually interested in."

chrisvire (approved):
@davidmoore1 approved.  Good job!

PR #836, opened 2025-06-12
Updated PlanStopDialog to Svelte 5

TheNonPirate:
Per #819

This PR updates PlanStopDialog.svelte to Svelte 5.

## Summary by CodeRabbit

- **Refactor**
	- Improved internal handling of dialog state and data binding for the plan stop dialog. No visible changes to appearance or functionality for end-users.

### DIFF
--- a/src/lib/components/PlanStopDialog.svelte
+++ b/src/lib/components/PlanStopDialog.svelte
@@ -2,7 +2,7 @@
 @component
 Plan Stop Modal Dialog component.
 -->
-<svelte:options accessors={true} />
+<svelte:options />
 
 <script>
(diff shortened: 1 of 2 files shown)


src/lib/components/PlanStopDialog.svelte
```diff
@@ -2,7 +2,6 @@
 @component
 Plan Stop Modal Dialog component.
 -->
-<svelte:options accessors={true} />
```
chrisvire, line 5: This needs to be replaced with $bindable and then `bind:planId` in the parent

chrisvire (approved):
👍

PR #848, opened 2025-06-22
Replace goto(getRoute()) by gotoRoute()

judah-sotomayor:
- Add the gotoRoute() function
- Replace all occurrences of goto(getRoute()) with gotoRoute() calls

This commit cleans up the codebase by reducing imports.
Files which had to import `goto` and `getRoute` from two different
locations can now simply import `gotoRoute`.

This also reduces the chance of error, e.g. developers forgetting to use `getRoute`.


chrisvire (approved):
Looks good!

Issue #492, opened 2024-05-07
Loading animation is still on the page after loading is done


chrisvire:
We have two different loading animations.  

One is while the website loads in `src/app.html`:

```
        <link rel="stylesheet" type="text/css" href="%sveltekit.assets%/spinner.css" />
```

And `src/lib/components/ScriptureViewSofria.svelte`:

```
<article class="container" bind:this={container}>
    {#if loading}
        <span class="spin" />
    {/if}
    <div
        id="content"
        bind:this={bookRoot}
        class:hidden={loading}
        style:font-family={font}
        style:font-size={fontSize}
        style:line-height={lineHeight}
        class="single"
        style:direction
    />
</article>
```

If you inspect the html, you can still see the spinner from app.html displayed under the other elements.  We should somehow remove that when we are done loading.  We added this since the animation from ScriptureViewSofria was taking a while before it was displaying. Surely we can do better.

PR #851, opened 2025-06-23
Removed loading animation after page loads

TheNonPirate:
Per #492

This PR causes the loading animation in app.html to be removed when the page finishes loading.
I did this by giving it an id and then using document.getElementById in ScriptureViewSofria and then just using .remove() to remove the element completely. I'm not sure if this is a bad practice. If it is, please tell me so I can find another way to solve the problem.

## Summary by CodeRabbit

- **New Features**
  - Added a fade-out animation to the loading spinner, which now disappears smoothly after the page loads.

- **Style**
  - Updated the loading spinner to support targeted animations and removal using a unique identifier.

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -1,17 +1,19 @@
 <!doctype html>
 <html lang="en">
-    <head>
-        <meta charset="utf-8" />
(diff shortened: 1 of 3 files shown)


chrisvire (approved):
👍

Issue #849, opened 2025-06-23
Styles/themes are not consistently managed across the application


judah-sotomayor:
Many elements use hardcoded color values.

<img width="1013" alt="Image" src="https://github.com/user-attachments/assets/1b4d79f9-73a4-4504-8aaf-4d341ab91131" />

This results in a number of style errors when changing between themes.

<img width="771" alt="Image" src="https://github.com/user-attachments/assets/72f7738d-cfba-43a9-be8b-1299aaf99cc7" />

<img width="541" alt="Image" src="https://github.com/user-attachments/assets/fdc873a3-b1eb-4f54-8f65-051de1c39cc8" />

<img width="541" alt="Image" src="https://github.com/user-attachments/assets/43e03575-1cf0-4e74-b94b-a1198e41d0d1" />

<img width="541" alt="Image" src="https://github.com/user-attachments/assets/6146c7b8-c9f8-4264-8b85-e84e36e43ba5" />

Icon colors, text color, primary and secondary foreground colors, background colors, etc. could all be refactored into a more consistent theming interface. Additionally, it would be nice to add a README documenting this interface for future developers to use while adding new components.

judah-sotomayor:
A possible solution to this could be setting daisyUI theme based on the SAB imported themes? This would enable use of the daisy elements pretty easily. https://daisyui.com/docs/colors/

I'm not sure if there are any problems with dynamically adjusting css variables. I'm also not sure how the import process works.

PR #856, opened 2025-06-26
Fix color theme inconsistencies.

judah-sotomayor:
Related to #849

Fix hard-coded color themes, and attempt to make colors more consistent across the application.


FyreByrd:
Would this be related to #984 @chrisvire ?

PR #645, opened 2024-07-19
Basic Storybook

gmjgeek:
Support storybooks with no audio.

Fixes #596 

# Implementation Notes

## Conversion

SAB tends to create storybooks 4-digit book codes of the form `B00#`. This is problematic for Proskomma, which only accepts 3-digit book codes. To resolve the issue, the PWA shortens these book codes to `B0#` when possible. If this new name would conflict with another book, the conversion process prints a warning to the console instead.

In storybooks, the conversion process tends to turn all formatting markers into [milestones](https://ubsicap.github.io/usfm/milestones/index.html). This approach helps keep storybooks compatible with Proskomma.

## Rendering

Unlike Scripture books, storybooks are not rendered by `ScriptureViewSofria.svelte`. Instead, they use `sofria-render-json/SofriaRender.svelte`. Here is the general process for rendering storybook text:

* Request the page as a Sofria JSON object from Proskomma
* Convert milestones and other features into a "render object" that is easier for Svelte to process
* Use a tree of Svelte components to convert the "render object" into HTML/CSS

Proskomma does not hold image data. Instead, illustrations for each storybook are listed in `config.js` and stored in the `illustrations` folder.

# TODO

## Features

- [X] Handle special storybook markers during content conversion
- [X] Display storybook illustrations at the top of each page
- [X] Render bold and italic texts
- [X] Render headings
- [X] Render ordered and unordered lists, including nested lists
- [X] Include boarder around storybook content
- [x] Navigate through pages like chapters
- [ ] Update convert process to be compatible with SAB 12.1
- [ ] When converting an app without storybooks, don't crash if the `data/illustrations` folder is not present
- [ ] Show only the illustration in landscape mode
- [ ] Make the illustration full-screen on double tap
- [ ] Keep boarder from scrolling when text overflows the screen (text should scroll within the boarder, which remains fixed)
- [ ] Ensure the boarder occupies the full width of the text area instead of shrinking to match content. Try creating a page with only one word. Does the boarder shrink, or does it retain its size?

## Bugs (12 August 2024):

- [ ] **Start with Page 1**: When navigating to a new storybook, the reader should land on the first page. Currently, if the user navigates from Matthew 5 to the "Unmerciful Servant" storybook, the app goes to page 5 of the storybook. If the storybook has fewer than 5 pages, the app crashes.
[...]

PR #870, opened 2025-07-21
#596 Basic Storybook Implementation

### LINKED ISSUE #596: Basic storybook - No audio
Storybooks are books that have a picture on each page and text that goes with it. In the more complex versions that will be covered in other issues, it can have audio playing, automatic page advancement, and Ken Burns type of motion on the images.  For the initial simplest case, PWA needs to support picture books without audio or motion.    Each page has a text and a picture defined.  The view changes depending upon whether the app is being viewed in landscape mode or portrait mode.  In portrait mode, the picture is displayed at the top of the screen and the text is viewed in the bottom of the screen.  In landscape mode, it only displays the picture as a full screen image.  

Portrait mode:
noaudioportrait

Landscape mode:
NoaudioLandscape

As with the scripture books, double tapping the screen changes it to full screen mode (and back)
noaudiolandscapefull

Advance from page to page as you would with the scripture paging.  Text for the first page of this book (which is displayed above)
```
\id 005
\toc2 No Audio

\page 1 
\img image1.jpeg
\m One day, Peter asked Jesus, “Master, how many times should I forgive my brother when he sins against me?  Up to seven times?” 
\m Jesus said, “Not seven times, but seventy times seven!” 
\m By this, Jesus meant that we should always forgive. Then Jesus told this story.
```
Besides the \img and \m markers, for img url and paragraph, the following entries may also be present:
\video followed by the video file name
\b blank line paragraph
\pc - centered paragraph
\pr - right justified paragraph

\s1 Heading 1
\s2 Heading 2
\s3 Heading 3

Configuration for a non audio storybook:
```xml
        <book id="005" type="story">
            <n>No Audio</n>
            <v>NA</v>
            <ct c="9" />
            <cn value="1-9" />
            <font-choice type="book-collection" />
            <f>UnfaithfulServant2.usfm</f>
            <audio-story-book mode="single-file" />
            <page num="1">
                <image-filename>UnfaithfulServant2-image1.jpeg</image-filename>
            </page>
            <page num="2">
                <image-filename>UnfaithfulServant2-image2.jpeg</image-filename>
            </page>
            <page num="3">
                <image-filename>UnfaithfulServant2-image3.jpeg</image-filename>
            </page>
            <page num="4">
                <image-filename>UnfaithfulServant2-image4.jpeg</image-filename>
            </page>
            <page num="5">
                <image-filename>UnfaithfulServant2-image5.jpeg</image-filename>
            </page>
            <page num="6">
                <image-filename>UnfaithfulServant2-image6.jpeg</image-filename>
            </page>
            <page num="7">
                <image-filename>UnfaithfulServant2-image7.jpeg</image-filename>
            </page>
            <page num="8">
                <image-filename>UnfaithfulServant2-image8.jpeg</image-filename>
            </page>
            <page num="9">
                <image-filename>UnfaithfulServant2-image9.jpeg</image-filename>
            </page>
            <features type="book">
                <e name="show-chapter-numbers" value="no" />
                <e name="lock-orientation" value="none" />
                <e name="show-border" value="inherit" />
                <e name="story-image-max-height" value="45" />
                <e name="show-in-book-selector" value="inherit" />
                <e name="show-verse-numbers" value="inherit" />
            </features>
            <images type="illustration">
                <image width="410" height="230">image1.jpeg</image>
                <image width="410" height="230">image6.jpeg</image>
                <image width="410" height="230">image7.jpeg</image>
                <image width="410" height="230">image8.jpeg</image>
                <image width="410" height="230">image4.jpeg</image>
                <image width="410" height="230">image5.jpeg</image>
                <image width="410" height="230">image9.jpeg</image>
                <image width="410" height="230">image2.jpeg</image>
                <image width="410" height="230">image3.jpeg</image>
            </images>
        </book>
```
Note that the size of the illustration may need to be adjusted by the "Resize the image files to reduct the app size" in Media/illustrations in SAB.  (features resize-illustrations-width if resize-illustrations is true).  This may also be overridden at the book level

TheNonPirate:
Per #596:
This PR is intended to modify storybook files into a form that the Proskomma USFM parser can parse. I took some code from #645.
I've gotten the basics of storybooks with text and images implemented.

## Summary by CodeRabbit

* **New Features**
  * Chapter-specific illustrations for story-type books shown above text with swipe navigation.
  * Support for styled inline spans, paragraph/character style milestones, and improved nested list rendering.
  * Book config now includes page illustrations metadata.

* **Bug Fixes**
  * Better handling of multi-file story books with correct chapter assembly and illustration mapping.

* **Tests**
  * Added Playwright E2E tests and unit tests for list transformations and storybook parsing.

* **Chores**
  * Test scripts updated to run Playwright + Vitest; Playwright added as a dev dependency.
  * .gitignore updated to exclude test-results.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sillsdev/appbuilder-pwa/pull/870?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


convert/convertBooks.ts
```diff
@@ -394,7 +402,6 @@ export async function convertBooks(
         for (const book of collection.books) {
             let bookConverted = false;
             switch (book.type) {
-                case 'story':
```
chrisvire, line 397: I know that it will fall to the default case, but I think it would be better to include `case 'story':` above `default:`

FyreByrd (changes requested):
Looks like a good start. Unfortunately, the pictures in the provided storybook can get quite large, which really messes with the text visibility.

Image

Image

On this largest one, I can't even scroll on the area where the text is supposed to be.
Image

TheNonPirate:
According to #596, it is not supposed to display the text when in landscape mode; it is only supposed to display the picture as a full screen image.

FyreByrd:
> According to #596, it is not supposed to display the text when in landscape mode; it is only supposed to display the picture as a full screen image.

Noted. I'll need to revisit this with Chris or David. I don't think this is the behavior we would want on laptop.

FyreByrd:
Per conversation with @davidmoore1 

1. In "portrait mode", the image should take up no more than half the screen's height.
2. The "portrait mode" vs. "landscape mode" feature was designed for mobile phones.
3. Users generally aren't going to be flipping a laptop on it's side.
4. Users may be using the PWA on a phone, so we still want to support the expected phone behavior.
5. Laptop users could have a button to switch between landscape and portrait mode. This could also be part of the double tap instead...

It is also worth noting that what you have right now is not actually a proper implementation of "landscape mode"

TheNonPirate:
What would a proper implementation of landscape mode be?

FyreByrd:
> What would a proper implementation of landscape mode be?

See the second image in https://github.com/sillsdev/appbuilder-pwa/issues/596#issue-2344112295

I think what's missing here is that in landscape mode the picture is supposed to be deliberately fullscreen.

There is still some active discussion between me, Chris, and David about how we want to handle landscape vs. portrait mode for the PWA, so this can wait for a bit until we've reached an agreement in our discussion.

convert/runTests.ts
FyreByrd, line 1: What is this for?

src/routes/text/+page.svelte
```diff
@@ -60,6 +60,10 @@
     import { onDestroy, onMount } from 'svelte';
     import { pinch, swipe } from 'svelte-gestures';
 
+    const illustrationURLs = import.meta.glob('$assets/illustrations/*', { eager: true });
```
FyreByrd: This should follow the exact same pattern as borders below.

src/routes/text/+page.svelte
```diff
@@ -495,59 +516,74 @@
...
-                                <ScriptureViewSofria {...viewSettings} />
-                            {/if}
-                        </div>
-                    </main>
+    <div class="flex flex-col overflow-y-auto">
+        {#if bookType === 'story'}
+            <!-- svelte-ignore a11y_missing_attribute -->
+            <img
+                src={getCurrentIllustrationFile()}
+                class="w-screen object-cover"
```
FyreByrd: This should have a responsive breakpoint that limits its size on larger screens.

FyreByrd (changes requested):
1. This needs to be rebased on main.
2. Based on conversation with Chris, we don't need to support "landscape" mode. This also means that the storybook image should have a maximum width set on larger screens.
3. I think it would be good to add the playwright testing as a separate PR, especially since nothing else uses playwright yet.
4. For some reason (and this *may* be out of scope?) I'm not seeing the arrows to navigate between chapters.
5. Also, the storybook included in the test project does not include *any* of the additional USFM tags that conversion steps and tests were added for...

chrisvire:
https://github.com/sillsdev/appbuilder-pwa/blob/e0bc61707ec23f49943ec67a3dcb5e191e310d05/src/routes/text/%2Bpage.svelte#L549

The content on the pages should be limited to `md` width by `max-w-screen-md mx-auto`.  This way, not matter how wide the browser is, the content is limited to `md` width (https://tailscan.com/tailwind/sizing/max-w-screen-md) which is 768px

TheNonPirate:
It appears that the arrows don't appear for small width (probably due to the TailwindCSS classes applied on the divs that the arrows are on) on Scripture books, either, so I think that is an out-of-scope issue.
Does the width limitation I implemented look good, or should I do it differently?

FyreByrd (approved):
Looks good to me. I think this is good enough for the minimal implementation of storybooks. My original comment on the list conversion feature still stands, but I think we can address that by adding a second storybook with lists in the future when we implement storybook audio. The arrows aren't the most visible when they appear on top of the image, but that's probably fine and can be addressed later if needed.

PR #872, opened 2025-07-22
Add w-full to contents center div.

judah-sotomayor:
Previously the central div in the contents tab looked pretty bad. This fixes it.


chrisvire (approved):
👍

Issue #862, opened 2025-07-07
Bug: clicking on bookmark/note/highlight with contents menu


chrisvire:
You can view the bookmarks, notes, and highlights from the hamburger menu. Normally, when you click on one of those, it goes to that item.

I was testing with the SAB Test PWA project:
https://drive.google.com/file/d/1cCo_W4Mi3itQn8VVSWioOsXcQ05Mj3jK/view?usp=drive_link

When I click on an item, it goes to the contents menu instead.

PR #874, opened 2025-07-23
Fix hyperlinks in highlight and bookmark cards.

judah-sotomayor:
Per #862

Previously clicking a highlight or a bookmark would jump the app to the contents page.
This commit corrects the href to point at `#/text` instead.

This commit also removes the ColorCard component, which is nearly duplicate with IconCard.
The aesthetic of ColorCard is kept by using the icon snippet to produce a color square in the highlights page.

## Summary by CodeRabbit

* **New Features**
  * Added support for dynamic links in highlight cards, allowing for customizable navigation.

* **Refactor**
  * Replaced the previous highlight card component with an updated card design that uses icons and supports dynamic link destinations.

* **Chores**
  * Removed the old highlight card component to streamline the codebase.


chrisvire (approved):
👍

PR #884, opened 2025-07-30
---

judah-sotomayor:
title: Update Lexicon to Svelte 5
draft: true
---

This PR introduces two related changes.
1. Update the Lexicon system to use Svelte 5 runes instead of stores, and simplify the state system.
2. Update Lexicon components and pages to Svelte 5 runes and remove old Svelte 4 reactivity.


chrisvire:
@judah-sotomayor What is the status of this PR? I would like @FyreByrd to finish it up.

judah-sotomayor:
All code should be up now.

PR #893, opened 2025-08-06
Update Plans to Svelte 5

chrisvire:


davidmoore1 (approved):
I have opened a separate issue about the scrolling being incorrect, but everything in the actual plan pages seemed correct.  The references are being set correctly, so I think the problems are in the text pages.

Issue #865, opened 2025-07-08
Break files out of cache with hashes


chrisvire:
Files need to have a hash in the filename so that when updates to the app are published, then the new files are used and the older versions will be aged out of the cache. Using a stable hash will make sure that if the file doesn't change, the previous cached version will be used. Realistically, anything that is written to the static folder should have a hash included in the filename.

We will need to update any references to the files (e.g config.json or collection/index.json) so that they can be found at runtime. We may need to introduce new files to track these names where paths were assumed before.

You may be able to use Node.JS's  built-in md5 hash?
https://stackoverflow.com/a/57915489

It would be good to "clean" the directory of older files (ones that don't match the hash but have the same base filename). It may be easier to delete all files an regenerate. There are some files that checked into static.  We could move those to a separate folder (like static-base?) that gets copied in after cleaning the folder.

PR #895, opened 2025-08-08
Implement hashed file paths

judah-sotomayor:
Per #865.

This pull request will implement hash-based access for asset files that are designed to be updated.
This includes most images, language resources, and other assets that do not need direct access.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,6 @@ static/watermarks
 static/borders
 static/clips
 static/videos
-static/icons
 static/*.partial.html
 static/manifest.json
 static/assets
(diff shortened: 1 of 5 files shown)


judah-sotomayor:
@chrisvire this currently breaks linting, because we cannot build using imports that are not being unpacked.
There are two options I can see to resolve this.
1. Make asset imports conditional on something. This doesn't really make sense for a lot of the codebase, because many images and assets are _not_ optional. However, it does keep our workflow light.
2. Add an unpack phase to the lint step that extracts one of the test projects and runs the converter scripts. This will take more workflow time, but it ensures our builds fully work every time.

PR #916, opened 2025-12-02
Fix ordering of words with special chars

FyreByrd:
Words with smart quotes were being sorted ahead of words without. As far as I can tell, this problem only affected reversals.

Questions:
- Are smart quotes the only characters we should be worried about?

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed sorting of lexicon entries to properly handle special characters, including smart quotes
  * Enhanced word ordering to maintain language context through improved locale-aware comparison

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/convert/convertReverseIndex.ts
+++ b/convert/convertReverseIndex.ts
@@ -72,7 +72,9 @@ export function convertReverseIndex(
 
     Object.entries(entriesByLetter).forEach(([letter, entries]) => {
         const fileLetter = letter.toLowerCase();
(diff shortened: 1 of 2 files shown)


FyreByrd:
Based on discussion with @chrisvire and an independent investigation of the app-builders codebase, it should be enough for now to ignore only the smart quotes for sorting reversal entries during the conversion process. In the future, we may also need to rework how we do reversals in general to more closely align with the Android app.

PR #917, opened 2025-12-02
Use AppBuilders Version in key for project_data cache

FyreByrd:
Tests were broken when app-builders added some new features. This does not fix those tests. This allows broken tests to be caught earlier by making sure that the latest version of app-builders is used to extract the project data for testing. Some of this was delayed by the fact that we were caching the extracted data to save time during the testing process. Using the version number as part of the cache key bypasses this issue.

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflow to include version detection and caching optimization for improved build consistency and reliability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,6 +10,9 @@ jobs:
     setup:
         runs-on: ubuntu-latest
 
+        outputs:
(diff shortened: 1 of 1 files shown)


chrisvire (approved):
Any caching of project generation should be based on the appbuilder_verison

Issue #948, opened 2026-03-18
Footnotes not scrolling correctly on Safari (mobile and desktop).


chrisvire:
See this post:
https://community.scripture.software.sil.org/t/sab-mac-13-3-2-pwa-display-notes/6951

Using the first note on Gen 1:1 on https://evangelischebijbelvertaling.nl/pwas/#/text

On Chrome, it is fine:

<img width="903" height="297" alt="Image" src="https://github.com/user-attachments/assets/d9c81440-c788-4ebf-a28b-236098808296" />

On safari, it is not displaying correctly:

<img width="848" height="301" alt="Image" src="https://github.com/user-attachments/assets/f8267b8b-b3b4-4b81-b243-20ef3ee74469" />

chrisvire:
Fixed #949

PR #949, opened 2026-03-19
Issue #948 - Footnotes not scrolling properly

davidmoore1:
Footnotes that needed to be scrolled did not display properly on Safari.   They showed blank areas under the initially displayed text when scrolled.  On a phone, when I scrolled up and back it would eventually paint the text correctly in Safari.

Changed drop-shadow-lg to shadow-lg

According to chatGPT:
drop-shadow-lg uses filter: drop-shadow(...)
In safari:
filter + overrflow + dynamic content = known repaint bug
Especially inside:
- flex layouts (which we have)
- positioned elements (absolute, bottom-8)
- scrolling containers (overflow-y-auto)

So the recommendation was to use shadow-lg unless we really needed it.

## Summary by CodeRabbit

* **Style**
  * Updated shadow effects on dropdown menus and stack view components for improved visual consistency.

### DIFF
--- a/src/lib/components/Dropdown.svelte
+++ b/src/lib/components/Dropdown.svelte
@@ -46,7 +46,7 @@ A simple dropdown menu from DaisyUI.
     <!-- svelte-ignore a11y_no_noninteractive_tabindex -->
     <div
         bind:this={container}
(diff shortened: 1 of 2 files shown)


chrisvire (approved):
👍

PR #953, opened 2026-05-04
Incorporate feedback from DLS into DAB PWA

FyreByrd:
Changes:
- fix navbar size
- consolidate lexicon components (most were completely unnecessary)
- fix sort order discrepancy with native
- fix entry images
- improve reversal loading
- move loading to layout (prevents navigation to/from search page causing reload)
- fix missing home button
- fix text size missing

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve line breaks in About page content.

* **Style**
  * Updated navbar layout and spacing.
  * Redesigned search form checkboxes and sidebar item styling.

* **New Features**
  * New lexicon full-screen layout with integrated search, inline language/letter navigation, and unified word list.
  * Improved entry rendering: better image/audio handling and consistent homonym subscript display.

* **Refactor**
  * Centralized lexicon state/navigation and unified text-appearance visibility logic.


src/lib/lexicon/components/EntryView.svelte
```diff
@@ -62,7 +62,8 @@
 
         function processNode(node: Node, parentHasSenseNumber = false) {
             if (node.nodeType === Node.TEXT_NODE) {
-                return node.nodeValue.trim();
+                // TODO: Data in DB has long runs of spaces...
+                return node.nodeValue.trim().replaceAll(/ +/g, ' ');
```
eomerdws: Are you seeing a performance hit with trimming a lot of white space on each load? Is this something we should consider doing in the conversion?

src/lib/lexicon/components/EntryView.svelte
```diff
@@ -45,7 +45,11 @@
         if (!xmlString) return '';
 
         const parser = new DOMParser();
-        const xmlDoc = parser.parseFromString(xmlString, 'text/xml');
+        // TODO: find better solution for <default font> replacement (this was causing xml parse errors)
+        const xmlDoc = parser.parseFromString(
+            xmlString.replaceAll("'<default font>',", ''),
```
eomerdws, line 66: Does this not have an ending tag? The space in the tag is also throwing me off.

src/lib/lexicon/components/EntryView.svelte
FyreByrd: Not really? This only runs when rendering an entry. Running something like this during the conversion process would require interacting with the SQLite database, which we aren't really doing during the conversion process.

src/lib/lexicon/components/EntryView.svelte
FyreByrd: I probably should have made it a bit more clear what was happening here. This is a string that was showing up in the inline `style="..."` attribute in some of the XHTML entries. I think it may be a bug in FieldWorks?

chrisvire: I sent a Slack message to @jasonleenaylor and @FyreByrd to talk about it.

eomerdws (dismissed):
Looks like awesome work overall. While checking data I did randomly click on a word that failed with an XML parsing error. I went through maybe till the r's and found 3 words total that seem to have this error. See screenshot.

From the console: XML parsing error: This page contains the following errors:error on line 1 at column 340: StartTag: invalid element name
Below is a rendering of the page up to the first error.

The words that failed for me are:

- Asteur
- Babiche
- Chaud


Image

Note: Steps I took (just in case I misstepped or missed something)
- I set the PWA location under Settings | Developer to where I checked out your PR. 
- I then used Build PWA Files.
- Ran npm ci
- Ran npm run dev

Not as important, but I noticed that there are a lot of blank entries at the beginning of the PWA that are not in the app. Probably for another PR.

Image

FyreByrd:
> Not as important, but I noticed that there are a lot of blank entries at the beginning of the PWA that are not in the app. Probably for another PR.

The blank entries are caused by DAB and FieldWorks. The necessary changes in DAB are made here: https://github.com/sillsdev/app-builders/pull/2214

FyreByrd:
@coderabbitai review

FyreByrd:
The remaining parse errors appear to be out of scope for the PR. I have opened an issue to resolve the underlying problem directly in DAB: https://github.com/sillsdev/app-builders/issues/2219

eomerdws (approved):
Looks like awesome work overall. The previous problems appear to be in DAB rather than in the PWA.

PR #869, opened 2025-07-10
Feature/integrate eslint

judah-sotomayor:
# Reconfigure ESLint to work well with prettier
Per #864.
This pull-request includes several related changes:
1. Reconfigure ESLint using a modern configuration and good rules for svelte.
   These rules include svelte and typescript extensions,
   as well as an integration to disable rules that conflict with prettier.
   All configuration is now consolidated in `eslint.config.js`, eliminating `.eslintrc.cjs`.
2. Disable noisy rules. A number of rules are loud/unnecessary.
   I've turned off `no-dupe-style-properties`, `require-each-key`, and `no-unused-svelte-ignore` because they are spurious for our codebase.
   I've also disabled the typescript `no-require-imports` for the tailwind configuration file, which uses `require` style imports.
3. Correct low-level eslint warnings.
   While I haven't fixed some of the more in-depth issues, every trivial issue should be corrected in this series of commits.
   This includes a number of empty objects, missing blocks around case statements, and a hundred or so conversions from `var` and `let` to `const`.
   As best as I can tell no behavior changes were introduced here.
   I am a bit unsure about 8a67ec3a, which changes the types on the CatalogData interface to `object` from the empty-object type. This could cause an issue if empty-object behavior was intended, but that seems unlikely.

As it stands, the linting tests will fail because of seven unresolved errors.
Each of these is more in-depth than I want to do in a short commit, and will receive a separate PR after some discussion about the best way to resolve them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected a typo in the inline style for font display in the font list.
  * Fixed test assertions to properly invoke matcher functions, ensuring accurate test validations.
  * Improved property existence checks for safer object handling in theme and store utilities.

* **Style**
  * Standardized variable declarations to use `const` and `let` appropriately across many files for improved immutability and clarity.
  * Added curly braces to case blocks in switch statements and single-line conditionals for consistent block scoping and style.
  * Updated inline style and prop value syntax for consistency and correctness.

* **Refactor**
  * Broadened type annotations for catalog data to increase flexibility.
  * Updated parameter types in analytics logging for consistency.
  * Enhanced ESLint configuration structure and updated related dependencies.

* **Chores**
  * Upgraded and added several dependencies, including ESLint, TypeScript, and Svelte-related packages.
  * Removed obsolete ESLint configuration file and replaced it with a modernized config.

* **Tests**
  * Improved test assertions for clarity and correctness.
  * Strengthened validation logic in test cases for book chapters.

[...]

Issue #864, opened 2025-07-08
Use ESLint and Prettier together to format code


chrisvire:
We currently use Prettier to format code.  However, it doesn't enforce some rules:
  https://chatgpt.com/share/686d7315-6364-8006-a85d-1d38f4c9400f

> Prettier does not support enforcing curly braces for single-line if statements. Its design philosophy is opinionated and intentionally avoids offering configuration for stylistic preferences that don't affect readability or parsing (like optional braces).

This GPT conversation suggests changes to use eslint and prettier together for formatting.  See if this works.

FyreByrd:
Fixed by #957

PR #957, opened 2026-05-13
Reconfigure ESLint to work well with prettier

FyreByrd:
Rebase of #869 to resolve merge conflicts

PR comment from #869:
Per #864.
This pull-request includes several related changes:
1. Reconfigure ESLint using a modern configuration and good rules for svelte.
   These rules include svelte and typescript extensions,
   as well as an integration to disable rules that conflict with prettier.
   All configuration is now consolidated in `eslint.config.js`, eliminating `.eslintrc.cjs`.
2. Disable noisy rules. A number of rules are loud/unnecessary.
   I've turned off `no-dupe-style-properties`, `require-each-key`, and `no-unused-svelte-ignore` because they are spurious for our codebase.
   I've also disabled the typescript `no-require-imports` for the tailwind configuration file, which uses `require` style imports.
3. Correct low-level eslint warnings.
   While I haven't fixed some of the more in-depth issues, every trivial issue should be corrected in this series of commits.
   This includes a number of empty objects, missing blocks around case statements, and a hundred or so conversions from `var` and `let` to `const`.
   As best as I can tell no behavior changes were introduced here.
   I am a bit unsure about 8a67ec3a, which changes the types on the CatalogData interface to `object` from the empty-object type. This could cause an issue if empty-object behavior was intended, but that seems unlikely.

As it stands, the linting tests will fail because of seven unresolved errors.
Each of these is more in-depth than I want to do in a short commit, and will receive a separate PR after some discussion about the best way to resolve them.

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Fixed font selection styling that was not applying due to an attribute typo.

* **Code Quality & Improvements**
  * Enhanced error handling and defensive checks across data validation and conversions.
  * Improved type safety with better TypeScript typing and null/undefined guards.
  * Upgraded linting configuration for stricter code standards and consistency.
  * Updated development dependencies for improved tooling support.


FyreByrd (commented):
I'm not seeing the seven remaining errors that are referred to. There aren't really any substantive changes between this and the original PR. This is mostly a rebase to resolve merge conflicts. It won't let me approve it myself, since I am now technically the PR author.

PR #959, opened 2026-05-14
Convert generated config objects to TS

FyreByrd:
Changes:
- moved generated config.js to gen-assets
- Export config as `AppConfig` type
- Include exports for `ScriptureConfig` and `DictionaryConfig`
- Similar changes for firebase-config.js and contents.js

## Summary by CodeRabbit

* **Refactor**
  * App now reads consolidated generated TypeScript configuration and content assets; many UI components and stores updated to use the new runtime config.

* **Style**
  * Stronger typing for feature/config schemas and content types; program fields made required for consistency.

* **Bug Fixes**
  * Improved resilience and null-safe handling for missing or partial configuration and content data to prevent runtime errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sillsdev/appbuilder-pwa/pull/959)

Issue #905, opened 2025-11-18
Use TypeScript instead of JavaScript


FyreByrd:
I have encountered multiple runtime errors that likely would have been prevented if we had been using TS across the entire codebase. We should make this move sooner rather than later. We should probably do this *before* converting to Svelte 5 (#819), or perhaps in tandem with those changes.

FyreByrd:
Commands that should be helpful in searching:
`git grep "<script>"`
`git ls-tree --full-tree --name-only -r HEAD | grep ".js$"`

FyreByrd:
There are a few files that probably should not be converted to TS, notably:
postcss.config.cjs and tailwind.config.cjs (will be picked up by #951)
service-worker.js
app-builder-video.js

PR #960, opened 2026-05-19
Convert stores and page loads to TS

FyreByrd:
Continuation of #959, part 2/3 of #905 

Changes:
- convert all stores to TS
- convert all page loads to TS
- convert icon exports to TS
- remove unused scripts
  - `groupStore`
  - old proskomma query.js and render.js

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed responsive layout in notes editor—divider and reference display now visible on smaller screens.

* **Refactor**
  * Modernized modal system with type-safe enum-based identifiers.
  * Enhanced TypeScript type annotations across data stores, database handlers, and page loaders.
  * Consolidated localization module to TypeScript with improved type safety.
  * Improved audio system typing and internal state management.


src/routes/plans/+page.ts
```diff
@@ -1,14 +1,11 @@
...
     // Make annotations more resiliant to changes in book collection ids over time.
     // Ignore entries if we don't currently have a collection for the entry.
     // This can happen during testing of different PWA at the same port and we don't
     // want to break the feature.
-    const plans = [];
-    depends('plans');
-    return { plans };
-}
+    depends('note:plans');
+    return { plans: [] as PlanItem[] };
```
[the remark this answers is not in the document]
FyreByrd: I'm not going to fix this here. The current plans page does not use any of the data returned from the page load for some reason. I appreciate the concern, but this is out of scope for the PR.

src/lib/data/stores/scripture.ts
```diff
@@ -197,30 +199,39 @@ export const fontChoices = derived(refs, ($refs) => {
     return currentFonts;
 });
 
+export type Selection = {
```
eomerdws, line 202: For chapters and verses why not make them a number?

src/lib/data/stores/audio.ts
```diff
@@ -23,27 +23,57 @@ function createaudioHighlightElements() {
...
     playStart: 0
-};
+} as AudioPlayer;
 export const audioPlayer = writable({ ...audioPlayerDefault });
 
-export const PLAYMODE_CONTINUE = 'continue';
-export const PLAYMODE_STOP = 'stop';
-export const PLAYMODE_REPEAT_PAGE = 'repeatPage';
-export const PLAYMODE_REPEAT_SELECTION = 'repeatSelection';
+export const PlayMode = {
```
eomerdws, line 60: I really like this approach over the old approach.

eomerdws (approved):
This is really good work. My comments aren't blocking an approval review, but one is a question for consideration and the other is I like the approach you took in this area. In fact you did it in at least one other place, but I didn't want you to have to resolve it multiple times.

src/lib/data/stores/audio.ts
FyreByrd: Yeah. I like it as well. Technically I could have done it as an `enum`, but I've found through my work on Scriptoria that `enum` is not the greatest in TS.

src/lib/data/stores/scripture.ts
FyreByrd: I'd have to check, but I think it's because the chapter and verse number are stored as a string somewhere else? Maybe in IndexedDB? Maybe as `data-*` attributes? I don't quite remember. I personally think `number` would be a better type too, but I gave it a type that reflected current usage. I may end up refactoring it in the future. My goal with this PR was to minimize substantive changes.

PR #399, opened 2023-08-21
Feature/verse on image

Caleb-Parks:

Issue #300, opened 2023-06-28
Implement Verse on Image feature


chrisvire:

chrisvire:
There were two issues:
* SAB was generating the watermark images to the root folder instead of a watermarks subfolder (like backgrounds) -- fixed in SAB 10.4 which should be out today or tomorrow
* convert process was copying files from watermarks folder -- fixed in #302

chrisvire:
For creating videos with ffmpeg in the browser, see; https://stackoverflow.com/a/65746732

PR #963, opened 2026-05-22
Feature/verse on image

TheNonPirate:
Implementing #300 and building on #399. This implements the verse on image feature. Currently, everything is implemented except for the video download option.

## Release Notes

* **New Features**
  * Enabled text-on-image editing with adjustable text styling (font, size, bold, italic, alignment, letter spacing, line height, color, and text effects).
  * Added image manipulation controls including brightness, contrast, saturation, and blur adjustments.
  * Introduced image upload and interactive cropping workflow.
  * Added share and download functionality for edited images.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sillsdev/appbuilder-pwa/pull/963?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


TheNonPirate:
Some of it still needs to be updated to Svelte 5 as well.

static/firebase-config.js
FyreByrd, line 1: This should not have been committed. Since this has API keys in it, please make sure to remove it via interactive rebase so that it doesn't show up in the git history.

FyreByrd: It looks like all the files that should not have been committed took place in the same commit, so it should work to just remove that one.

src/routes/image/+page.js
FyreByrd, line 1: If you're not going to return anything from the page load, you shouldn't need this script at all.

src/routes/crop/+page.js
FyreByrd, line 1: If you're not going to return anything from the page load, you shouldn't need this script at all.

src/lib/components/DownloadSelector.svelte
FyreByrd, line 1: Please convert this to use TypeScript

src/lib/components/VerseOnImage.svelte
FyreByrd, line 1: Please convert this to use TypeScript

package.json
```diff
@@ -59,7 +59,9 @@
         "proskomma-tools": "^0.0.5",
         "rimraf": "^6",
         "sharp": "^0.34.5",
+        "simple-color-picker": "^1.0.5",
```
FyreByrd: What is this dependency for? I'm not seeing it used anywhere

src/lib/components/VerseOnImage.svelte
FyreByrd, line 1: This has so many variables that are unnecessarily prefixed with `voi_`. If namespacing is a concern, what I'd recommend is to combine all of the variables into one object with named properties. You could then also codify this as a specific type that could be extracted and used elsewhere if needed?

FyreByrd (changes requested):
Well. The code itself works, and it works pretty well. I was able to generate an image following largely the same process that is present in the native app.

However, there are definitely quite a few design issues, most of which were inherited from the original PR, so I wouldn't feel too bad about it.

Here's what I noticed in addition to the individual review comments:

1. The image page could use a little bit of padding/margin on the top and bottom.
2. I used a really large image on the image upload page `/crop`. It may be good to have a maximum size for the image, at least when rendering to the page? It may also be beneficial to display the image dimensions somewhere on the page (or just the selection's dimensions)
3. I think it would be helpful to disable native text selection within the image in the verse on image page. (I found it really annoying when trying to drag the text around that it kept trying to highlight).

Any additional thoughts @chrisvire ? I know you had said you had some issues with the original PR.

src/lib/components/VerseOnImage.svelte
TheNonPirate: I think I've gotten it pretty well converted. Is there anything else you want me to change on the TypeScript front?

TheNonPirate:
@FyreByrd Is this goal with the limitation on image size just to prevent the user from uploading an image that will take a really long time to render? It doesn't seem like uploading a larger image causes the preview on the screen to be any bigger.

src/lib/components/VerseOnImage.svelte
FyreByrd: Several variables are shown as being possibly null/undefined and are showing errors in the editor.

src/lib/components/VerseOnImage.svelte
```diff
@@ -0,0 +1,1072 @@
...
+        style="
+            overflow-x: scroll;
+            overflow-y: visible;
+            min-height: 2.7rem;
+            z-index: 3;
+            --tabWidth: {imageWidth / 5}px; 
+            background-color: {$themeColors['ImageTabsBackgroundColor']};
+        "
+    >
+        <!-- NavBar of tab buttons to bring up the different editor panes -->
```
FyreByrd, line 627: This could probably be simplified into an `#each` loop?

src/lib/icons/image/BlurIcon.svelte
FyreByrd, line 1: The extra type annotation shouldn't be necessary? Same comment for all the other icons.

src/routes/image/upload/+page.svelte
FyreByrd, line 1: Probably bikeshedding a bit here, but I'm not a big fan of the route naming. What about moving it under `/image` and renaming it to `upload` (so `/image/upload`)

FyreByrd (changes requested):
Shouldn't be too much left to change. Just what I have here and also the CR comments.

src/lib/icons/image/BlurIcon.svelte
TheNonPirate: Without the extra type annotation that included the ? indicating values were optional, I was getting TypeScript errors when just the color or size property (Rather than both) was being used.

FyreByrd: I'm not getting that error... I guess it's fine to leave it in...

FyreByrd (approved):
Looks good to me! All of my feedback has been addressed. Thanks for the good work.

PR #965, opened 2026-06-01
#489 Navigation properly hiding and displaying for empty chapters based on user selection

### LINKED ISSUE #489: Navigation: do not display entries for missing chapters
The code that generates the `catalog.js` handles the case if the chapter tag is missing in the book file.  However, it is still including the chapter if there are actually no verses.

This change should affect the GraphQL code so that chapters are not included in the catalog if there are no verses in the chapter.

@davidmoore1 will provide more details.

eomerdws:
Fix for users selecting "Hide empty chapters" in SAB, but the PWA still displays them. 

Also ensured that unchecking "Hide empty chapters" also shows those missing chapters in the navigation.

The initial issue is https://github.com/sillsdev/appbuilder-pwa/issues/489.

## Summary by CodeRabbit

* **New Features**
  * Added an option to hide empty chapters in book and chapter selection interfaces.

* **Bug Fixes**
  * Fixed an issue where disabling the hide-empty option could result in missing chapters; selection now shows chapters correctly when the option is off.

### DIFF
--- a/src/lib/components/BookSelector.svelte
+++ b/src/lib/components/BookSelector.svelte
@@ -261,7 +261,7 @@ The navbar component.
                         .filter(
                             (y) =>
                                 typeof chapters[y] === 'object' &&
-                                Object.keys(chapters[y]).length > 1
+                                Object.keys(chapters[y]).length > 0
                         )
                         .map((x) => ({
                             label: getChapterLabel(x),
--- a/src/lib/components/BookSelector.svelte
+++ b/src/lib/components/BookSelector.svelte
@@ -341,7 +341,7 @@ The navbar component.
                   bookLabel: 'abbreviation'
               })
             : bcv === c
-              ? chapterGridGroup(chapters)
+              ? chapterGridGroup()
               : verseGridGroup(chapter)}
         {menuaction}
     />
--- a/src/lib/components/BookSelector.svelte
+++ b/src/lib/components/BookSelector.svelte
@@ -248,9 +248,9 @@ The navbar component.
         return groups;
     };
 
-    let chapterGridGroup = (chapters: Record<string, unknown>) => {
+    let chapterGridGroup = () => {
         let hasIntroduction = books.find((x) => x.bookCode === book)?.hasIntroduction;
-
+        console.log(chapters);
         if (hideEmptyChapters) {
             return [
                 {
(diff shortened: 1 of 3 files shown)


chrisvire:
Please follow this naming convention for PR/commit title:
https://github.com/naming-convention/naming-convention-guides/tree/master/git

src/lib/components/BookSelector.svelte
```diff
@@ -249,17 +250,38 @@ The navbar component.
...
+        if (hideEmptyChapters) {
+            return [
+                {
+                    rows: hasIntroduction
+                        ? [{ label: $t['Chapter_Introduction_Title'], id: 'i' }]
+                        : null,
+                    cells: Object.keys(chapters)
+                        .filter(
+                            (y) =>
+                                typeof chapters[y] === 'object' &&
```
FyreByrd: What is this check meant to accomplish? `typeof null` also equals `'object'`, so depending on what is intended, this may end up with some errors.

src/lib/components/ChapterSelector.svelte
```diff
@@ -179,10 +180,18 @@ The navbar component.
                                 }
                             ]
                           : undefined,
-                      cells: Object.keys(chapters).map((x) => ({
-                          label: getChapterLabel(x),
-                          id: x
-                      }))
+                      cells: hideEmptyChapters
```
FyreByrd: Realistically, this could be simplified by checking `hideEmptyChapters` in the filter condition instead. I know this would technically be bad code from a theoretical performance standpoint, but we know that the largest input it will ever have to handle has length of 150 (Psalms). Same comment applies to BookSelector.

src/lib/components/ChapterSelector.svelte
```diff
@@ -179,10 +180,18 @@ The navbar component.
...
+                          ? Object.keys(chapters)
+                                .filter(
+                                    (y) =>
+                                        typeof chapters[y] === 'object' &&
```
FyreByrd: We already know what type `chapters` and its constituents are, so I don't think this really needs to be checked.

src/lib/components/BookSelector.svelte
```diff
@@ -249,17 +250,38 @@ The navbar component.
 
     let chapterGridGroup = (chapters: Record<string, unknown>) => {
```
FyreByrd: This is only ever called in one place, with the parameter being the `$derived` `chapters` variable. We can probably remove this as a function parameter. This would also simplify some of the typing. This is partially my fault.

FyreByrd (changes requested):

src/lib/components/ChapterSelector.svelte
eomerdws: That's a great point it would simplify the code and you are right nothing too big would come along.

eomerdws:
I *think* I've made all the requested changes.

FyreByrd (approved):

PR #967, opened 2026-06-01
Implement Video Download Button for Verse On Image

TheNonPirate:
Implementing #300 and continuing my work from #963. This implements the video download feature for verse on image. When audio files are available for the chapter, clicking the download button causes a small menu to pop up with "Save Image" and "Save Video" options. When "Save Video" is clicked, it creates and downloads a video consisting of a still frame of the image and text, with the audio playing the audio associated with the selected verse(s). If the browser can support it, it creates the video as an mp4 file. Firefox (At least on my machine) does not support the audio encoding needed for mp4, so for any browsers that do not support it, it downloads it as a webm file instead.

## Summary by CodeRabbit

* **New Features**
  * Export and download verses as single-frame video files with integrated audio when available.
  * Active “Save video” option in the download modal; page button now triggers video export or immediate image download based on audio presence.
  * Video export displays a progress overlay with cancel support; outputs saved as WebM or MP4 depending on codec support.

* **Chores**
  * Added a media runtime dependency to support the new export workflow.

### DIFF
--- a/src/lib/components/VerseOnImage.svelte
+++ b/src/lib/components/VerseOnImage.svelte
@@ -16,6 +16,7 @@ The verse on image component.
         refs,
         s,
         selectedVerses,
+        t,
         themeColors,
         voiCustomImage,
         windowSize
@@ -1242,8 +1243,8 @@ The verse on image component.
               flex flex-col justify-between h-64"
         >
             <div class="flex flex-col gap-1 text-left">
-                <p class="text-lg font-bold">Save Video</p>
-                <p class="text-sm">Creating video...</p>
+                <p class="text-lg font-bold">{$t['Text_On_Image_Save_Video']}</p>
+                <p class="text-sm">{$t['Video_Creating_Video']}</p>
             </div>
 
             <div class="w-full">
@@ -1255,7 +1256,7 @@ The verse on image component.
                     class="dy-btn dy-btn-sm dy-btn-ghost"
                     onclick={() => {
                         cancelDownload = true;
-                    }}>Cancel</button
+                    }}>{$t['Button_Cancel']}</button
                 >
             </div>
         </div>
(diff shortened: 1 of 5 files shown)


src/lib/components/VerseOnImage.svelte
```diff
@@ -427,6 +442,212 @@ The verse on image component.
...
+                timestamp: 0,
+                duration: 1_000_000
+            });
+            sample = new VideoSample(frame);
+
+            const audioConfig: AudioEncodingConfig = await pickSupportedAudioConfig();
+
+            const output = new Output({
+                format:
+                    audioConfig.codec === 'opus' ? new WebMOutputFormat() : new Mp4OutputFormat(),
```
FyreByrd: This condition is checked against multiple times. It could be good to separate out into its own variable?

FyreByrd (approved):
The code appears to work fine. I'm ok merging as is. The process *might* be able to be simplified a bit by doing the audio encoding in the background while the user is still editing the image. We could then, perhaps, disable the video download option if the audio encoding fails? Thoughts @chrisvire ?

TheNonPirate:
One thing that I've noticed with the verse on image feature is that when the download modal is open, it shifts the image slightly to the left on certain browsers (At least Chromium ones). It seems that this happens with all of the modals because DaisyUI applies a scrollbar-gutter to the page when a model is opened (https://daisyui.com/components/modal/). However, the text is not shifted because it uses absolute positioning, so it slightly messes up the position of the text on the image for the export. The easiest way to fix it would be to exclude rootscrollgutter, but that would affect every modal in the PWA, which may or may not be desirable. I'm not sure what another good way to fix it would be.

FyreByrd:
> The easiest way to fix it would be to exclude rootscrollgutter, but that would affect every modal in the PWA, which may or may not be desirable. I'm not sure what another good way to fix it would be.

We do eventually want to do this, but we won't be able to until after we upgrade DaisyUI.

PR #881, opened 2025-07-30
Feature - Songbooks

KarsonDahl:
* Create new component for Songbook

* Add new route for Songbooks

* Create svelte page for individual songs and songbook menu

* Borrow icon until icon can be selected

Very beginning steps- work still needed in all four of these areas. Proper layout configuration for the Songbooks menu, a fully fleshed-out component, proper implementation of the data for each individual Songbook, as well as a permanent icon.

PR #972, opened 2026-06-03
Implement songbooks

### LINKED ISSUE #859: Support Song Books
In the Building Apps document, section 17, there is documentation for Song Books.  They can be added to a collection and have a special display which shows an index by number and alphabetical.

TODO: Add screenshots from SAB and an App

TheNonPirate:
Closes #859 and building on #881. This implements songbooks. When a songbook is selected, the app goes to a page that allows the user to select which song they want, sorted by number or title. Currently, the song's text is rendered the same as any scripture book, which means that it does not perfectly match the mobile app.

## Release Notes

* **New Features**
  * Added songs as a supported book type with a dedicated interface
  * Two browsable views for songs: browse by title or by number
  * Search capability enabled for songs collections

* **Changes**
  * Verse selection automatically hides when viewing songs content


convert/convertBooks.ts
```diff
@@ -611,6 +611,9 @@ function convertHtmlBook(context: ConvertBookContext, book: BookConfig, files: a
 
     let content = fs.readFileSync(srcFile, 'utf-8');
     const before = getHashedNameFromContents(content, book.file);
+    if (book.type === 'songs') {
```
chrisvire: @TheNonPirate I don't think this is the correct place to do this.  This isn't an HTML Book.

TheNonPirate: Oops

src/lib/components/ChapterSelector.svelte
```diff
@@ -18,7 +18,12 @@ The navbar component.
     // Needs testing, does updating the book correctly effect what chapters or verses are availible in the next tab?
     const book = $derived($nextRef.book === '' ? $refs.book : $nextRef.book);
     const chapter = $derived($nextRef.chapter === '' ? $refs.chapter : $nextRef.chapter);
-    const showVerseSelector = $derived($userSettingsOrDefault['verse-selection']) as boolean;
+    const showVerseSelector = $derived(
+        $userSettingsOrDefault['verse-selection'] &&
+            scriptureConfig.bookCollections
```
FyreByrd: A similar filter should be in BookSelector?

src/lib/components/ChapterSelector.svelte
```diff
@@ -18,7 +18,12 @@ The navbar component.
...
+                ?.find((x) => $refs.collection === x.id)
+                ?.books.find((x) => book === x.id)?.type !== 'songs'
+    ) as boolean;
```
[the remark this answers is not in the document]
FyreByrd: @TheNonPirate if you use `$derived.by` you can have a slightly more complicated derivation (i.e. you can have one with an internal variable (e.g. `const book = <complicated lookup>`)).

src/lib/components/Sidebar.svelte
```diff
@@ -77,6 +78,7 @@ The sidebar/drawer.
     const showNotes = !!config.mainFeatures['annotation-notes'];
     const showHighlights = !!config.mainFeatures['annotation-highlights'];
     const showPlans = scriptureConfig.plans?.plans.length;
+    const showSongbooks = true;
```
FyreByrd: This appears to be incomplete.

src/lib/icons/SongBookIcon.svelte
FyreByrd, line 1: This looks like an "info" icon. Is this what we want?

Image

src/lib/types.ts
```diff
@@ -7,3 +7,9 @@ export interface SearchFormSubmitEvent {
     wholeWords: boolean;
     matchAccents: boolean;
 }
+
+export interface SongBook {
```
FyreByrd: This does not appear to be used anywhere...

src/routes/songs/[collection]/[id]/+page.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (!numberUrl) {
+            throw new Error(`Song text asset not found for key ${numberKey}`);
+        }
+        const numberResponse = await fetch(numberUrl);
+        if (!numberResponse.ok) {
+            throw new Error('Failed to fetch song text file');
+        }
+
+        numberData = await numberResponse.text();
+    } catch (error) {
```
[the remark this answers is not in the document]
Comment (FyreByrd) [reply src/routes/songs/[collection]/[id]/+page.ts:35]: @TheNonPirate I second this feedback. Everything else in this file looks fine.

src/routes/songs/[collection]/[id]/+page.svelte
```diff
@@ -0,0 +1,103 @@
...
+    import { navigateToText } from '$lib/navigate';
+    import type { PageData } from './$types.js';
+
+    interface Props {
+        data: PageData;
+    }
+    let { data }: Props = $props();
+
+    let songsByTitle: { number: string; title: string }[] = [];
+    data.titleData.split('\r\n').forEach((value) => {
```
Comment (FyreByrd) [inline src/routes/songs/[collection]/[id]/+page.svelte]: This data processing should either be a `$derived` or it should be done in the page load instead. Same comment for `numberData`.

FyreByrd (changes requested):

src/routes/songs/[collection]/[id]/+page.svelte
```diff
@@ -0,0 +1,103 @@
...
+    let borderColor = $derived($s?.['ui.song.border']['background-color']);
+    const songArray = $derived(curTab === 0 ? songsByNumber : songsByTitle);
+
+    async function completeNavigation(songNumber: string) {
+        await navigateToText({
+            docSet: $refs.docSet,
+            collection: $refs.collection,
+            book: $refs.book,
+            chapter: songNumber,
+            verse: $nextRef.verse
```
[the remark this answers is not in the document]
Comment (TheNonPirate) [reply src/routes/songs/[collection]/[id]/+page.svelte:36]: @FyreByrd Is this really a potential issue? I wouldn't think that $refs.collection or $refs.book could have problems with this, but I don't know for sure.

src/lib/components/ChapterSelector.svelte
TheNonPirate: The behavior in the native app is that the verse selector still shows up for the book selector, but not for the chapter selector. Do we want to keep it how it is to match the native app, or change it because it doesn't make much sense to still have the verse selector for songs?

TheNonPirate:
I realized that SAB has a "Hide verse number 1 if it is directly after the chapter number" feature, which is enabled in the project that I'm testing but can be disabled. This means that my approach of inserting a verse number 1 at the beginning of the song in order to display the chapter number won't work because it can lead to the dummy verse number being displayed.

src/lib/components/Sidebar.svelte
TheNonPirate: This (And the SongBookIcon) was part of some code in the previous attempt at songbooks. It added an icon in the menu that apparently could be used to go directly to the songbook. However, I can't find any setting in SAB that would cause such an icon to show up. Do you know of any such setting, or should I just remove this completely?

FyreByrd (changes requested):
The indentation on the first line doesn't look right compared to the rest of the lines.
Image
Native for comparison:
Image

We may also want to horizontally center the container??

Also, Hark the Herald Angels Sing is missing a lot of the first letter of each line, this should probably be fixed.

Finally, there *are* some merge conflicts, so you'll need to rebase again.

src/lib/components/Sidebar.svelte
FyreByrd: I have no idea. That would be a question for @chrisvire

src/lib/components/ChapterSelector.svelte
FyreByrd: In my mind at least, it doesn't really make any sense to keep it, but on the other hand fidelity to the native app is still important... Thoughts @chrisvire ?

src/routes/songs/[collection]/[id]/+page.svelte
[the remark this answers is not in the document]
Comment (FyreByrd) [reply src/routes/songs/[collection]/[id]/+page.svelte:36]: I think this should be fine as is. I wouldn't worry about it.

TheNonPirate:
I think the indentation issue is there on Scripture books as well, just not as obviously generally. In this example, the same general issue can be seen:
image
image

I'm pretty sure the container is also not horizontally centered in Scripture books either.
Should these be fixed in this PR or a separate one?

src/lib/components/Sidebar.svelte
TheNonPirate: I decided to go ahead and remove it. If it really is a setting that we need to support, I can reimpliment it.

src/lib/components/ChapterSelector.svelte
```diff
@@ -18,7 +18,16 @@ The navbar component.
     // Needs testing, does updating the book correctly effect what chapters or verses are availible in the next tab?
     const book = $derived($nextRef.book === '' ? $refs.book : $nextRef.book);
     const chapter = $derived($nextRef.chapter === '' ? $refs.chapter : $nextRef.chapter);
-    const showVerseSelector = $derived($userSettingsOrDefault['verse-selection']) as boolean;
+    const showVerseSelector = $derived.by(() => {
+        const bookType = scriptureConfig.bookCollections
+            ?.find((x) => $refs.collection === x.id)
+            ?.books.find((x) => book === x.id)?.type;
+        if (!bookType) {
```
FyreByrd: If you're checking if bookType is truthy here, then why are you checking it again just below?

src/routes/songs/[collection]/[id]/+page.ts
```diff
@@ -0,0 +1,78 @@
...
+        console.error('Error fetching song number text file:', error);
+    }
+    const songsByTitle: { number: string; title: string }[] = [];
+    const songsByNumber: { number: string; title: string }[] = [];
+    titleData.split(/\r?\n/).forEach((value) => {
```
Comment (FyreByrd) [inline src/routes/songs/[collection]/[id]/+page.ts]: This could probably be simplified into a `map` followed by a `filter` and could also be relocated to happen in the return. Same comment applies to songsByNumber.

src/routes/songs/[collection]/[id]/+page.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+    numberData.split(/\r?\n/).forEach((value) => {
+        const separatedValue = value.split('\t');
+        if (separatedValue.length === 2) {
+            songsByNumber.push({ number: separatedValue[0], title: separatedValue[1] });
+        }
+    });
+
+    return {
+        titleData: titleData,
```
Comment (FyreByrd) [inline src/routes/songs/[collection]/[id]/+page.ts]: You don't need to return this, since the processed version that you actually use in the front end is already being returned. Same comment for numberData

src/routes/songs/[collection]/[id]/+page.ts
```diff
@@ -0,0 +1,78 @@
...
+        numberData: numberData,
+        songsByTitle: songsByTitle,
+        songsByNumber: songsByNumber,
+        collection: params.collection,
```
Comment (FyreByrd) [inline src/routes/songs/[collection]/[id]/+page.ts]: You aren't using this variable in the front-end. Same comment on songId and displayLabel...

src/routes/songs/[collection]/[id]/+page.svelte
```diff
@@ -0,0 +1,91 @@
...
+    import { SearchIcon } from '$lib/icons';
+    import { navigateToText } from '$lib/navigate';
+    import type { PageData } from './$types.js';
+
+    interface Props {
+        data: PageData;
+    }
+    let { data }: Props = $props();
+
+    let songsByTitle = $derived(data.songsByTitle);
```
Comment (FyreByrd) [inline src/routes/songs/[collection]/[id]/+page.svelte]: I don't think you need this to be a separate `$derived`. Accessing it directly from `data` in the one place it's used should be fine. Same comment on songsByNumber.

src/routes/songs/[collection]/[id]/+page.svelte
```diff
@@ -0,0 +1,91 @@
...
+                    {songTab}
+                </button>
+            {/each}
+        </div>
+        {#each songArray as song}
+            <button
+                class="w-full block justify-start text-left pl-10 border"
+                style:border-color={borderColor}
+                onclick={() => completeNavigation(song.number)}
+                ><span style={convertStyle($s?.['ui.song.number'])}>{song.number}</span>
```
Comment (FyreByrd) [inline src/routes/songs/[collection]/[id]/+page.svelte]: The elements here are a bit cluttered. I don't know why prettier won't automatically take care of it, but if you add new lines between the button tags and their contents, then run prettier, it should be fine.

FyreByrd (changes requested):
This is really close. Just a few nitpicks at this point. I think we may want to wait until #986 is merged before this one is.

TheNonPirate:
I've updated it.

src/routes/songs/[collection]/[id]/+page.ts
```diff
@@ -51,28 +44,25 @@ export const load: PageLoad = async ({ params, fetch }) => {
...
     return {
-        titleData: titleData,
-        numberData: numberData,
-        songsByTitle: songsByTitle,
-        songsByNumber: songsByNumber,
-        collection: params.collection,
-        songId: id,
-        displayLabel: book?.name || 'Song'
+        songsByTitle: titleData
+            .split(/\r?\n/)
```
Comment (FyreByrd) [inline src/routes/songs/[collection]/[id]/+page.ts:41]: This could be a bit cleaner. You don't actually have to make this change, but you do have merge conflicts to deal with.

```ts
titleData
    .split(/\r?\n/)
    .map((line) => line.split('\t'))
    .filter((parts) => parts.length === 2)
    .map((parts) => ({ number: parts[0], title: parts[1] }))
```

FyreByrd (changes requested):
One last nitpick. You don't *have* to make that change, but you do still have a merge conflict. Once the conflict is resolved I'm good to approve this.

TheNonPirate:
This ended up being more complicated than I expected. I ended up going ahead and hiding the verse selector in the BookSelector component as well as in the ChapterSelector. That isn't exactly what the native app does, but it makes more sense, and the native app displayed the verse as "i" while the PWA was displaying it as a number or as no number at all, so it wasn't matching the native app anyway. I also found some problems relating to how I navigated to the song that caused the chapter number on the top of the screen to sometimes not update when moving between chapters.
It also appears that I didn't even fully resolve the conflicts, so I'll get back to work on those.

TheNonPirate:
It should be good now, but enough did change that it's probably worth another review.

src/lib/components/BookSelector.svelte
```diff
@@ -42,9 +42,25 @@ The navbar component.
 
     const showChapterSelector = config.mainFeatures['show-chapter-selector-after-book'] as boolean;
     const listView = $derived($userSettingsOrDefault['book-selection'] === 'list');
-    const showVerseSelector = $derived($userSettingsOrDefault['verse-selection'] as boolean);
+    const showVerseSelector = $derived.by(() => {
+        const bookType = scriptureConfig.bookCollections
+            ?.find((x) => $refs.collection === x.id)
+            ?.books.find((x) => book === x.id)?.type;
+        if (!bookType) {
```
FyreByrd: 49-52 could be simplified down to `return $userSettingsOrDefualt['verse-selection'] && (!bookType || bookType !== 'songs');`

Similar comment on 59-62 and the same in ChapterSelector.

FyreByrd: Realistically, these could possibly be separated out into function calls with module scope in BookSelector, then imported into ChapterSelector.

```html
// BookSelector.svelte
<script module lang="ts">
  export function hideEmptyChapters(book: string): boolean {
    const bookType = scriptureConfig.bookCollections
        ?.find((x) => $refs.collection === x.id)
        ?.books.find((x) => book === x.id)?.type;
    return $userSettingsOrDefault['hide-empty-chapters'] && (!bookType || bookType !== 'songs');
  }
</script>

<script lang="ts">
  ...
  const _hideEmptyChapters = $derived(hideEmptyChapters(book));
  ...
</script>

// ChapterSelector.svelte
<script lang="ts">
  ...
  import { hideEmptyChapters } from '$lib/components/BookSelector.svelte';
  ...
  const _hideEmptyChapters = $derived(hideEmptyChapters(book));
  ...
</script>
```

src/lib/components/ChapterSelector.svelte
FyreByrd, line 1: See comment in BookSelector

src/routes/songs/[collection]/[id]/+page.ts
```diff
@@ -0,0 +1,60 @@
...
+    query: '?url'
+});
+
+export const load: PageLoad = async ({ params, fetch }) => {
+    const id = params.id;
+    const collection = params.collection;
+
+    let titleData = '';
+    let numberData = '';
+    try {
```
Comment (FyreByrd) [inline src/routes/songs/[collection]/[id]/+page.ts:16]: Sorry to change this up on you so much, but I think this could be simplified down to:
```ts
try {
    const titleKey = `./${collection}-${id}-songs-by-title.txt`;
    const titleUrl = songs[titleKey];
    if (!titleUrl) {
       console.error(`Song text asset not found for key ${titleKey}`);
    } else {
        titleData = await fetch(titleUrl).then((r) => r.text());
    }
} catch (error) {
    console.error('Error fetching song title text file:', error);
}
```

Same comment for numbers.

FyreByrd (changes requested):
Almost there. Just some nitpicks

TheNonPirate:
Does it look good now?

FyreByrd (approved):
For the future, unless there are merge conflicts with the base branch (which I can't verify after the fact) or a feature in the base branch that you need for the PR (which I am fairly certain you didn't), please refrain from rebasing and force-pushing. Except in the aforementioned circumstances, it is really unnecessary and clogs up the PR history, making it harder to see what changes you've made after specific comments.

PR #973, opened 2026-06-04
#505 Finish implementing "repeat selection" audio mode

### LINKED ISSUE #505: Repeat button in Selection Bar doesn't work
The button is present in the selection toolbar that appears when you select text.  Pressing the button should result in it cycling through the different repeat modes, similar to what it does on the audio bar.  At the moment, I don't see anything happen when I press the repeat mode button.  It doesn't change.  I couldn't find a way to get it to play just the selection or to repeat on the selected text.

benpaj-cps:
Fixes #505.

Upon investigating #505 I found that the functionality for playing an audio selection on repeat was incomplete. The audio state machine has already been written, but it didn't support configuring a playback range based on the first and last selected verses (like the Android app does).

- Fixed an existing typo that prevented "repeat selection" mode from activating
- Added an onclick handler for "repeat selected audio" button in `TextSelectionToolbar`
- Wrote logic in `data/audio.ts` for determining the correct playback range from the first and last selected verses.
- Debugged and improved code quality in `data/audio.ts`

I tried to match the native Android behavior closely. The only significant difference that I am aware of is that the native version waits until the end of the current phrase to jump back to the beginning of the selection if the user attempts to seek beyond the end of the selection, while in this PR the jump-back is immediate.

Simple testing on an installed PWA on Android 16 looks good so far. The following are desktop screenshots. (Note: the "repeat selection" hint only shows when changing modes using the mode buttton, not the repeat selection button. If this is not intended behavior I can look at addressing that.)

image

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted audio play-mode reset to trigger only when timing data is truly available.

* **Improvements**
  * Updated text selection playback so Play starts from the first selected verse, and Repeat plays the full selected verse range.
  * Improved play/pause, stop, and seek behavior—better handling when audio isn’t loaded and more consistent pause/resume.
  * Refined verse timing calculations and playback range fallbacks for smoother listening updates.

### DIFF
--- a/src/lib/components/AudioBar.svelte
+++ b/src/lib/components/AudioBar.svelte
@@ -100,7 +100,7 @@ TODO:
     const iconPlayColor = $derived($s?.['ui.bar.audio.play.icon']['color']);
     const backgroundColor = $derived($s?.['ui.bar.audio']['background-color']);
     const audioBarClass = $derived($refs.hasAudio?.timingFile ? 'audio-bar' : 'audio-bar-progress');
-    $effect(() => mayResetPlayMode($refs.hasAudio?.timingFile));
+    $effect(() => mayResetPlayMode(!!$refs.hasAudio?.timingFile));
     $effect(() => updatePlaybackSpeed($userSettings['audio-speed'] as string));
 </script>
 
@@ -116,8 +116,8 @@ TODO:
     {#if showRepeatMode}
         <div class="audio-control-buttons">
             <RepeatButton
-                color={iconColor}
-                onclick={() => playMode.next($refs.hasAudio?.timingFile)}
+                color="{iconColor},"
+                onclick={() => playMode.next(!!$refs.hasAudio?.timingFile)}
                 state={$playMode.mode}
             />
         </div>
--- a/src/lib/components/AudioBar.svelte
+++ b/src/lib/components/AudioBar.svelte
@@ -36,6 +36,7 @@ TODO:
         // If the current mode is repeatSelection and the reference is changed to something without timing
         // (even chapter without audio), then reset the playMode.  This matches how the Android app behaves.
         if (!hasTiming && $playMode.mode === 'repeatSelection') {
+            console.log(`mayResetPlayMode: calling reset()`);
             playMode.reset();
         }
     }
@@ -100,7 +101,11 @@ TODO:
     const iconPlayColor = $derived($s?.['ui.bar.audio.play.icon']['color']);
     const backgroundColor = $derived($s?.['ui.bar.audio']['background-color']);
     const audioBarClass = $derived($refs.hasAudio?.timingFile ? 'audio-bar' : 'audio-bar-progress');
-    $effect(() => mayResetPlayMode($refs.hasAudio?.timing));
+    $effect(() => mayResetPlayMode($refs.hasAudio?.timingFile));
+    // $effect(() => {
+    //     console.log(`effect: thinks that timing is ${$refs.hasAudio?.timing}`);
+    //     mayResetPlayMode($refs.hasAudio?.timing);
+    // });
     $effect(() => updatePlaybackSpeed($userSettings['audio-speed'] as string));
 </script>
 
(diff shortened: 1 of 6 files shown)


benpaj-cps:
[This](https://github.com/sillsdev/appbuilder-pwa/pull/973/changes#diff-aac9dd9f95c7b0a9c71a2590e41e7fdc12c50cc3a2900e57f7e5aba26761199dL103) also fixes the issue that the audio mode button could not be set to "repeat selection" because of a conditional check which always evaluated to false and reset the mode to "continue" (`hasAudio` has no attribute called `timing`).

benpaj-cps:
(from CodeRabbit above)
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > src/lib/data/audio.ts (1)> `291-309`: _⚠️ Potential issue_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Edge case: undefined check doesn't guard against zero-valued ranges.**
> > > The check on lines 294-295 tests for `undefined`, but `getCurrentVerseTiming()` always returns numeric values (defaulting to 0, not undefined). If timing data is missing or malformed, this creates a range `{start: 0, end: 0}`, and the end-of-range check on line 304 will immediately trigger warmdown when progress > 0.
> > > Consider checking for a zero `end` value instead:
> > > 
> > > Suggested fix
> > > ```diff
> > >              const resultArray = getCurrentVerseTiming();
> > > -            if (resultArray?.start === undefined || resultArray.end === undefined) {
> > > +            if (resultArray.end === 0) {
> > > +                // Invalid timing range - don't set it
> > >                  return;
> > >              }
> > > ```

This isn't exactly accurate. This code block will only execute if the `range` hasn't been set already (i.e. is at the default value of `{ start: 0, end: 0 }`. Setting the range to zero in the edge case that `getCurrentVerseTiming` doesn't return meaningful data is no more than a no-op.

If anything this could be an additional error case, but I'm not sure if that's needed as this logic pre-dates the current PR.

benpaj-cps:
@chrisvire Ready for review. I can add more details on the updated logic early next week if needed.

src/lib/components/TextSelectionToolbar.svelte
```diff
@@ -170,12 +175,18 @@ TODO:
                 </div>
             {:else}
                 {#if isAudioPlayable && $refs.hasAudio && $refs.hasAudio.timingFile}
```
FyreByrd: Please convert this to use optional chaining. Same comment on line 185.

src/lib/components/AudioBar.svelte
```diff
@@ -100,7 +100,7 @@ TODO:
     const iconPlayColor = $derived($s?.['ui.bar.audio.play.icon']['color']);
     const backgroundColor = $derived($s?.['ui.bar.audio']['background-color']);
     const audioBarClass = $derived($refs.hasAudio?.timingFile ? 'audio-bar' : 'audio-bar-progress');
-    $effect(() => mayResetPlayMode($refs.hasAudio?.timing));
+    $effect(() => mayResetPlayMode($refs.hasAudio?.timingFile));
```
FyreByrd: This happened because `hasAudio` is of type `any`.  It would be beneficial to fix the type issue, so this mistake is able to be caught more easily. This may be out of scope for the PR, so you don't have to do it, but it is a good opportunity to contribute to our efforts to eventually fix all of the type issues.

FyreByrd (approved):
Works well. Just one small nitpick, and a larger request that you are fully welcome to ignore. Approving the PR so you are free to fix these if you so choose.

src/lib/components/TextSelectionToolbar.svelte
benpaj-cps: Fixed.

src/lib/components/AudioBar.svelte
benpaj-cps: Ok. I attempted a fix for this, let me know what you think. (after I fix the lint errors...)

src/lib/data/stores/reference.ts
```diff
@@ -14,7 +15,7 @@ interface ReferenceStore extends Reference {
     collection: string;
     chapterVerses: string;
     numVerses: number;
-    hasAudio: any;
+    hasAudio: BookCollectionAudioConfig | undefined;
```
FyreByrd, line 18: For future reference, you could also do this as `hasAudio?: BookCollectionAudioConfig`

FyreByrd (approved):

Issue #980, opened 2026-06-08
\b USFM tag is not rendered by ScriptureViewSofria


TheNonPirate:
One USFM tag mentioned in the USFM documentation and used in some SAB projects is \b (https://ubsicap.github.io/usfm/poetry/index.html). This is supposed to insert a blank line. However, it currently does not have any effect.

This is a screenshot from the PWA:

<img width="729" height="939" alt="Image" src="https://github.com/user-attachments/assets/21b98a1f-0cd7-4835-8c1a-25232ff2ad6a" />

This is a screenshot from the native app:

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/33d005e2-141e-4286-9167-17422f7a5d20" />

After verse 11, there is a \b tag. In the native app, it creates a line break, but in the PWA, it does not.

PR #986, opened 2026-06-10
Render \b tags

TheNonPirate:
Implementing #980. This causes \b tags to properly be a line break.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scripture paragraph rendering: when a “main” paragraph resolves to the styling variant “b”, the view now adds an extra non-breaking space to prevent unintended spacing/layout issues and maintain consistent readability across scripture views.

### DIFF
--- a/src/lib/components/ScriptureViewSofria.svelte
+++ b/src/lib/components/ScriptureViewSofria.svelte
@@ -1649,6 +1649,7 @@ LOGGING:
                                             workspace.paragraphDiv.classList.add(paraClass);
                                             if (paraClass === 'b') {
                                                 workspace.paragraphDiv.innerHTML += '<br>';
+                                                workspace.paragraphDiv.classList.add('txs');
                                             }
                                         }
                                         break;
--- a/src/lib/components/ScriptureViewSofria.svelte
+++ b/src/lib/components/ScriptureViewSofria.svelte
@@ -1648,8 +1648,7 @@ LOGGING:
                                             workspace.paragraphDiv = document.createElement('div');
                                             workspace.paragraphDiv.classList.add(paraClass);
                                             if (paraClass === 'b') {
-                                                workspace.paragraphDiv.innerHTML += '<br>';
-                                                workspace.paragraphDiv.classList.add('txs');
+                                                workspace.paragraphDiv.innerHTML += '&nbsp;';
                                             }
                                         }
                                         break;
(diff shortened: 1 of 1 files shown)


benpaj-cps:
Looks good, but the `<b>` element is missing the `txs` class, which makes it smaller and harder to see.

image

image

The PWA has wider spacing than the Android version from the screenshots by default, so the default `<b>` is a bit too thin.

benpaj-cps (approved):
That did fix the spacing. Looks good to me.

benpaj-cps:
@chrisvire Looks like a staff approval is (understandably) still needed to merge PRs.

chrisvire:
@benpaj-cps You suggested adding the `txs` to the `<b>` tag.  Is that in done in the Native app (looking at the generated HTML for this section?

benpaj-cps:
> @benpaj-cps You suggested adding the `txs` to the `<b>` tag. Is that in done in the Native app (looking at the generated HTML for this section?

@chrisvire Turns out, no. Thanks for asking - I hadn't checked that. The Android app has an equivalent `&nbsp;` instead of the `<br>`.

Android:
image<br>

To match the Android version we could set the default line height to 1.2 and just use either `<br>` or `&nbsp;` without the `txs`.

Desktop PWA:
image

@TheNonPirate Sorry for the confusion!

TheNonPirate:
What do you mean by, "set the default line height to 1.2"? Are you suggesting some sort of global change, or applying a Tailwind class to that particular div?

benpaj-cps:
It's a UI adjustment, but the global default is set in lib/data/stores/view.ts:

image

I'm not sure if Chris wants to change this, but if so it should be a one-liner.

TheNonPirate:
It seems like changing a global default would go beyond the scope of this PR.
@chrisvire Thoughts?

chrisvire:
If you look at the HTML generated for the Native app, what styling is on the `<b/>` tag?

You can also right-click on the book in SAB and select Export to HTML and you can view it on the desktop.

benpaj-cps:
> If you look at the HTML generated for the Native app, what styling is on the `<b/>` tag?
> 
> You can also right-click on the book in SAB and select Export to HTML and you can view it on the desktop.

 The Android native app looks like it doesn't have a `<b/>` tag. It just has a single `&nbsp;` within the `<div class="b"/>`.
As far as I could tell the relevant styles in `sab-app.css` were identical.

(Android WebView CSS on left, desktop PWA on right)
image

TheNonPirate:
What should I adjust? Should I just give it the "b" class with &nbsp instead it?

benpaj-cps:
> What should I adjust? Should I just give it the "b" class with &nbsp instead it?

I'd suggest so, since that most closely matches the Android version. You're right that the line spacing default probably isn't an issue (at least not for this PR).

FyreByrd (approved):
Looks good to me. I love a simple one-line fix. (well, I guess it's three lines because of formatting, but close enough)

PR #988, opened 2026-06-10
Added swipe animations between chapters on PWA #230

### LINKED ISSUE #230: Swipe animation between chapters
At the moment when you swipe between chapters, you do not get any animation. 

It would be nicer to see something like in Global.Bible: https://myk.global.bible/
- which is like we see in our Android and iOS apps. You see the new page arriving and the old page going.

Awest882:
* Added a horizontal swiping motion between chapters
* Text for previous/next chapters will slide into the place of the
  current chapter
* Swipe handler activates when user swipes 1/3 of the text width
  (Adjustments can be made to `minSlideDistance`)
* Previous/Next buttons also trigger the animation
* Closes #230 

## New Features
- Added smooth horizontal mouse and drag navigation between chapters, with snapping and automatic transitions.
- Added animated previews of adjacent chapters for more fluid navigation.
- Selection changes now refresh chapter and book view settings reliably.

## Bug Fixes
- Improved drag tracking, movement limits, snapping, and overflow handling.
- Prevented verse selections from affecting unrelated scripture views.
- Improved verse selection and deselection reliability.

### DIFF
--- a/src/routes/text/+page.svelte
+++ b/src/routes/text/+page.svelte
@@ -678,6 +678,7 @@
 <svelte:window
     bind:innerWidth
     bind:innerHeight
+    // bind:visualViewport?.height={}
     onpointermove={handleMouseMove}
     onpointerup={handleMouseUp}
 />
@@ -806,10 +807,7 @@
                 />
             {/if}
         {/if}
-        <!-- <div class="overflow-y-auto grow" bind:this={scrollingDiv} onscroll={saveScrollPosition}> -->
         <div
-            // style="--borderImageSource: url({borders['./border.png']});"
-            // class:borderimg={showBorder}
             class="overflow-y-auto grow overflow-x-hidden"
             bind:this={scrollingDiv}
             onscroll={saveScrollPosition}
@@ -833,10 +831,14 @@
                     use:measure
                 >
                     <div
-                        class="p-2 w-full"
-                        style="position: absolute; left: {panels_X[0]}px; clip-path: inset(0 {1 *
-                            panels_X[0] +
-                            x.current}px 0 {-1 * panels_X[0] - x.current}px);"
+                        class="p-2 w-full overflow-y-hidden"
+                        style="position: absolute; left: {panels_X[0]}px;  height: {Math.abs(
+                            panels_X[0] + x.current
+                        ) === draggableWidth
+                            ? window.screen.height
+                            : 'auto'}px; clip-path: inset(0 {1 * panels_X[0] + x.current}px 0 {-1 *
+                            panels_X[0] -
+                            x.current}px);"
                     >
                         <main>
                             <div
@@ -865,10 +867,14 @@
                     </div>
 
                     <div
-                        class="p-2 w-full"
-                        style="position: absolute; left: {panels_X[1]}px;clip-path: inset(0 {1 *
-                            panels_X[1] +
-                            x.current}px 0 {-1 * panels_X[1] - x.current}px);"
+                        class="p-2 w-full overflow-y-hidden"
+                        style="position: absolute; left: {panels_X[1]}px;  height: {Math.abs(
+                            panels_X[1] + x.current
+                        ) === draggableWidth
+                            ? window.screen.height
+                            : 'auto'}px; clip-path: inset(0 {1 * panels_X[1] + x.current}px 0 {-1 *
+                            panels_X[1] -
+                            x.current}px);"
                     >
                         <main>
                             <div
--- a/src/routes/text/+page.svelte
+++ b/src/routes/text/+page.svelte
@@ -847,7 +847,10 @@
                 />
             {/if}
         {/if}
+        <!-- <div class="overflow-y-auto grow" bind:this={scrollingDiv} onscroll={saveScrollPosition}> -->
         <div
+            style="--borderImageSource: url({borders['./border.png']});"
+            class:borderimg={showBorder}
             class="overflow-y-auto grow overflow-x-hidden"
             bind:this={scrollingDiv}
             onscroll={saveScrollPosition}
--- a/src/routes/text/+page.svelte
+++ b/src/routes/text/+page.svelte
@@ -147,12 +147,11 @@
     let isDragging = $state(false);
     let draggableWidth = $state(0);
     let panels_X = $state([0, 0, 0]);
-    let minSlideDistance = () => draggableWidth / 3; // use to determine how far a user has to slide to move to the next chapter
-
-    /**
-     * Turn page previous <-  l - c - r =>              c - r - l(new page)
-     * Turn page next     ->  l - c - r =>    (new page)r - l - c
-     */
+    let minSlideDistance = () => draggableWidth / 2; // use to determine how far a user has to slide to move to the next chapter
+    let minSlideMomentum = 0.6; // measured in pixels per millisecond
+    let lastX = 0;
+    let lastTime = 0;
+    let momentum = 0;
 
     async function setupSettingsCache() {
         settingsCache[0] = {
@@ -225,9 +224,10 @@
         }
     }
 
-    async function handleMouseUp(_event: any) {
+    async function handleMouseUp(_event: PointerEvent) {
+        console.log(momentum);
         isDragging = false;
-        if (Math.abs(x.current) < minSlideDistance()) {
+        if (Math.abs(x.current) < minSlideDistance() && Math.abs(momentum) < minSlideMomentum) {
             x.set(0, { duration: Math.abs(x.current) });
             return;
         } else if (x.current < 0) {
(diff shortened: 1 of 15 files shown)


src/routes/text/+page.svelte
```diff
@@ -128,6 +132,75 @@
         }
     }
 
+    let x = new Tween(0);
+    let startX = 0;
+    let isDragging = $state(false);
+    let draggableWidth = $state(0);
+    let minSlideDistance = () => draggableWidth / 3; // use to determine how far a user has to slide to move to the next chapter
```
benpaj-cps: This is a bit confusing for the user. The watershed for which chapter to "fall" to should be at the midpoint between the chapters, not 33% percent of the way between them (see screen recording). However, the user also shouldn't need to swipe very far to trigger the transition.

I suggest using [a Spring](https://svelte.dev/docs/svelte/svelte-motion#Spring) instead, because it supports momentum - if the user releases early it could be set up to still trigger the transition, while the watershed can be set to 50% like it is on the [myk.global.bible](https://myk.global.bible) mobile version Chris mentioned.

src/routes/text/+page.svelte
```diff
@@ -128,6 +132,75 @@
...
+            x.set(0);
+            return;
+        } else if (x.current < 0) {
+            if (!(hasNext && navigateBetweenBooksNext)) {
+                x.set(0);
+                return;
+            }
+            await x.set(-draggableWidth);
+            await navigateToTextChapterInDirection(1);
+            x.set(0, { duration: 0 });
```
benpaj-cps: There is a slight visual jump whenever the transition completes and the chapter change occurs, even though this tween is set to be instant. Perhaps draggableWidth is off by a pixel or two?

src/routes/text/+page.svelte
```diff
@@ -179,7 +256,7 @@
         getFeatureValueString('audio-phrase-end-chars', $refs.collection, $refs.book)
     );
 
-    const showSearch = !!config.mainFeatures['search'];
+    const showSearch = !!config.mainFeatures['search']; // Why are there double negations on these??
```
benpaj-cps: I believe this is so that the computed result is always a boolean.

src/routes/text/+page.svelte
```diff
@@ -128,6 +132,75 @@
...
     }
 
+    let x = new Tween(0);
+    let startX = 0;
+    let isDragging = $state(false);
+    let draggableWidth = $state(0);
+    let minSlideDistance = () => draggableWidth / 3; // use to determine how far a user has to slide to move to the next chapter
+
+    async function handleMouseUp(_event: any) {
+        console.log('MOUSE UP');
```
benpaj-cps: (This and line 167 should be removed before merge.)

src/routes/text/+page.svelte
benpaj-cps: Maybe could use two springs - one until the transition midpoint and one after - the first moves toward the current chapter while the second moves toward the next chapter. Then the chapter navigation could be based on the current horizontal position of the windows, not the user's mouse or finger being released (i.e. the reference change would trigger only *if* the end of the horizontal motion was reached).

src/routes/text/+page.svelte
```diff
@@ -583,19 +687,46 @@
...
+                    use:measure
+                >
+                    <div
+                        class="p-2 w-full"
+                        style="position: absolute; left: {-draggableWidth}px; clip-path: inset(0 0 0 {draggableWidth -
+                            x.current}px);"
+                    >
+                        <main>
+                            <div class="max-w-screen-md mx-auto">
+                                {#if format === 'html'}
```
benpaj-cps: The previous chapter still shows even if it can't be navigated to (see video). This might possibly be outside this PR - I've noticed a similar issue of seeking audio backwards to navigate to a chapter that I can't move to directly.

benpaj-cps (commented):
I left a few comments. Overall it works reliably, but a few things might be frustrating for the end user. (this was as of yesterday's changes, so please disregard anything outdated).

https://github.com/user-attachments/assets/6493b179-a0bf-49b2-aecb-b755310e9279

Awest882:
To the recent comment, I just pulled in the changes from last night changing from Program Files (x86) to Program Files. Nothing should have changed for that.

src/routes/text/+page.svelte
Awest882: I've managed to get rid of the jump by reordering the Tween operations and chapter navigation, but now I've noticed that there is significant, noticeable overhead with loading/reloading the book views. I'm going to attempt to reduce this by only reloading one of the panels instead of all three.

if you have panels 0 - 1 - 2 and go to the left it will rotate the panels to be 1 - 2 - 0 and reload panel 0. Hopefully this will speed it up a little.

My laptop has no issues with the overhead, but my phone freezes up for a split second when it is loading the new content.

benpaj-cps:
Another thing I noticed was that borders aren't being handled properly. They aren't being shown at all (a different, existing issue), but the swipe animation doesn't involve the (blank) borders. Here's the Android version for comparison. The Android version also has its watershed set to maybe 60% of the way to the next chapter.


https://github.com/user-attachments/assets/dd89ac19-b738-4e5b-83d2-880fc9601fda


https://github.com/user-attachments/assets/23a45708-aea3-4118-bd0a-d6777732fcef

src/routes/text/+page.svelte
Awest882: It's a lot smoother now on transition. Also does not allow transition to the next book if the `navigateBetweenBooks` settings are disabled


https://github.com/user-attachments/assets/d1716ad9-1495-47e1-9f53-42e5694725f0

src/routes/text/+page.svelte
```diff
@@ -129,6 +132,157 @@
...
+            if (direction === -1) {
+                panels_X[panelX] = draggableWidth;
+            } else {
+                panels_X[panelX] = -draggableWidth;
+            }
+        }
+    }
+
+
+    async function handleMouseUp(_event: any) {
```
FyreByrd: Why not type this as `PointerEvent`? Same comment for other handlers.

FyreByrd (commented):
Switching chapters in the chapter selector is not triggering the animation (is it supposed to?). Also, switching chapters in the book selector should probably have the same behavior.

Text selection is also broken by this.

src/routes/text/+page.svelte
Awest882: I didn't know the type that svelte used for those events. It's been changed

Awest882:
> Switching chapters in the chapter selector is not triggering the animation (is it supposed to?). Also, switching chapters in the book selector should probably have the same behavior.

I wasn't planning on doing the animation because the selector could navigate to chapters in any direction and distance and thought it would be okay to just switch out the chapters.

I've been using myk.global.bible as a reference as mentioned in earlier comments, and it doesn't use an animation for the chapter selector

FyreByrd:
> I wasn't planning on doing the animation because the selector could navigate to chapters in any direction and distance and thought it would be okay to just switch out the chapters.

Understood. You probably shouldn't have any changes in ChapterSelector then??

Awest882:
> Understood. You probably shouldn't have any changes in ChapterSelector then??

Since I used three panels, not directly reliant on `viewSettings`, I couldn't find an easy way to get the updates from the chapter selector. I just added a custom event handler to notify the PWA when it was used. Otherwise the panels were not updated and were stuck at the previous chapter you were viewing

FyreByrd:
> Since I used three panels, not directly reliant on `viewSettings`, I couldn't find an easy way to get the updates from the chapter selector. I just added a custom event handler to notify the PWA when it was used. Otherwise the panels were not updated and were stuck at the previous chapter you were viewing

What data are you getting from the chapter selector that would not be accessible from the `$refs` store?

Awest882:
> What data are you getting from the chapter selector that would not be accessible from the `$refs` store?

I am just detecting when the action of selecting a chapter is complete. I tried using an $effect to see changes to `viewSettings`, but it would trigger whenever you click the right and left buttons or swipe. The purpose of this is to prevent the large overhead of loading in three pages of scripture at once. It must be done when using the selector because it doesn't cache the surrounding pages, but when navigating to the next or previous page it only loads in one page and shifts the other two over.

I just access the chapter selector and trigger the custom event when a new chapter is selected. Do you think there would be a better way for me to do this?

FyreByrd:
> I just access the chapter selector and trigger the custom event when a new chapter is selected. Do you think there would be a better way for me to do this?

That's actually a valid point. I wasn't thinking about the text loading. You could probably do something more along what I was asking for, but what you have now is probably good then. I guess just document it in a comment?

chrisvire:
> Switching chapters in the chapter selector is not triggering the animation (is it supposed to?). Also, switching chapters in the book selector should probably have the same behavior.
> 
> Text selection is also broken by this.

In the Android app, switching chapters in the chapter selector does not trigger the slide-to-side animation.  

If you select a verse in either the book or chapter selector (if verses are shown), then it loads the chapter without animation between chapters.  I have seen some inconsistent behavior while going to the version. I have see it animate to the verse (if it is offscreen) from verse 1 (with [Bhutan's Bible](https://play.google.com/store/apps/details?id=com.dzongkha.bible&hl=en_US)).  Another ([Haya Audio Bible](https://play.google.com/store/apps/details?id=com.haya.audio.bible&hl=en_US)), it jumps to the verse. However, the last time it was updated was last year.

You can enable "Show verse selector after selecting a chapter number" in Navigation > Book Navigation > Navigation page in SAB and test with current SAB to verify.  If we need the verse animation, we can create a separate issue for that.

chrisvire:
Finished in #1132

PR #992, opened 2026-06-11
Implement Audio Sharing for Selected Text

### LINKED ISSUE #377: TextSelectionToolbar: Share Audio

TheNonPirate:
Resolves #377. If there is audio timing data, this causes the share button when text is selected to open a modal that gives the user a choice between sharing text and sharing audio.

## Summary by CodeRabbit

* **New Features**
  * Added a Share dialog to share selected verses as text.
  * Added the ability to generate and share a trimmed chapter audio extract for the selected verses.

* **Improvements**
  * Integrated sharing into the app’s modal workflow, using native sharing when available and automatically falling back to download.
  * Improved audio sharing output selection and reliability, producing timing-accurate trimmed audio.

* **Bug Fixes**
  * Fixed the sharing flow to open the correct share modal and offer “text-only” sharing when verse timing data isn’t available.

### DIFF
--- a/src/lib/components/ShareSelector.svelte
+++ b/src/lib/components/ShareSelector.svelte
@@ -66,7 +66,6 @@ A component providing a dropdown where you can choose to download audio or video
             const audioSource = new AudioBufferSource(audioConfig);
             output.addAudioTrack(audioSource);
             await output.start();
-            console.log(audioSourceInfo?.source);
             let isRemote;
 
             try {
@@ -78,11 +77,12 @@ A component providing a dropdown where you can choose to download audio or video
             if (isRemote) {
                 try {
                     let verses = await selectedVerses.getCompositeText();
-                    navigator.share({
+                    await navigator.share({
                         title: reference,
                         text: verses,
                         url: audioSourceInfo.source
                     });
+                    return;
                 } catch (error) {
                     if ((error as { name?: string })?.name === 'AbortError') {
                         return; // user intentionally dismissed native share UI
--- a/src/lib/components/ShareSelector.svelte
+++ b/src/lib/components/ShareSelector.svelte
@@ -31,9 +31,12 @@ A component providing a dropdown where you can choose to download audio or video
         const bookCol = $selectedVerses[0].collection;
         const fullBook = getBook({ collection: bookCol, book: book });
         const bookAbbrev = fullBook?.abbreviation ?? fullBook?.name;
+        const copyShareMessage = scriptureConfig.bookCollections?.find(
+            (x) => x.id === bookCol
+        )?.copyShareMessage;
         shareText(
             scriptureConfig.name ?? '',
-            scriptureConfig.name + '\n\n' + text + '\n' + reference,
+            text + '\n' + reference + (copyShareMessage ? '\n' + copyShareMessage : ''),
             book + '.txt'
         );
         logShareContent('Text', bookCol, bookAbbrev ?? '', reference);
@@ -209,8 +212,12 @@ A component providing a dropdown where you can choose to download audio or video
     } //This is used to determine a supported audio configuration. It first tries AAC, but then falls back to opus if AAC isn't supported. This is a duplicate of the function with the same name in VerseOnImage.svelte, so maybe it should be moved to somewhere that exports it for any place that needs it to use it?
     let modalId = 'shareSelector';
     let modalThis: Modal;
-    export function showModal() {
-        modalThis.showModal();
+    export function showModal(shareTextOnly: boolean) {
+        if (shareTextOnly) {
+            shareSelectedText();
+        } else {
+            modalThis.showModal();
+        }
     }
     const positioningCSS = $derived(
         'position:absolute; bottom:' +
--- a/src/lib/components/ShareSelector.svelte
+++ b/src/lib/components/ShareSelector.svelte
@@ -1,0 +1,182 @@
+<!--
+@component
+A component for verse-on-image providing a dropdown where you can choose to download an image or a video
+-->
+
+<script lang="ts">
+    import { scriptureConfig } from '$assets/config';
+    import { getBook, logShareContent } from '$lib/data/analytics';
+    import { getAudioSourceInfo } from '$lib/data/audio';
+    import { shareText } from '$lib/data/share';
+    import { refs, selectedVerses, t } from '$lib/data/stores';
+    import { AudioIcon } from '$lib/icons';
+    import FormatAlignLeftIcon from '$lib/icons/image/FormatAlignLeftIcon.svelte';
+    import {
+        AudioBufferSource,
+        BufferTarget,
+        canEncodeAudio,
+        Mp4OutputFormat,
+        Output,
+        WebMOutputFormat
+    } from 'mediabunny';
+    import type { AudioEncodingConfig } from 'mediabunny';
+    import Modal from './Modal.svelte';
+
+    let { vertOffset = '1rem' } = $props();
+    const reference = $derived(selectedVerses.getCompositeReference());
+    async function shareSelectedText() {
+        const book = $selectedVerses[0].book;
+        const reference = selectedVerses.getCompositeReference();
+        const text = await selectedVerses.getCompositeText();
+        const bookCol = $selectedVerses[0].collection;
+        const fullBook = getBook({ collection: bookCol, book: book });
+        const bookAbbrev = fullBook?.abbreviation ?? fullBook?.name;
+        shareText(
+            scriptureConfig.name ?? '',
+            scriptureConfig.name + '\n\n' + text + '\n' + reference,
+            book + '.txt'
+        );
+        logShareContent('Text', bookCol, bookAbbrev ?? '', reference);
+    }
+    async function shareAudio() {
+        const audioCtx = new AudioContext();
+        try {
+            const audioConfig: AudioEncodingConfig = await pickSupportedAudioConfig();
+            const useWebm = audioConfig.codec === 'opus';
+            const output = new Output({
+                format: useWebm ? new WebMOutputFormat() : new Mp4OutputFormat(),
+                target: new BufferTarget()
+            });
+
+            const audioSourceInfo = await getAudioSourceInfo({
+                collection: $refs.collection,
+                book: $refs.book,
+                chapter: $refs.chapter
+            });
+            if (!audioSourceInfo?.source) {
+                throw new Error('No audio source available for this chapter');
+            }
+
+            const audioSource = new AudioBufferSource(audioConfig);
+            output.addAudioTrack(audioSource);
+            await output.start();
+
+            const audioBlob = await fetch(audioSourceInfo?.source).then((r) => r.blob());
+            const audioBuffer = await audioCtx.decodeAudioData(await audioBlob.arrayBuffer());
+
+            const sampleRate = audioBuffer.sampleRate;
+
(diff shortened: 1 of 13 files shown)


benpaj-cps:
On my phone (Pixel, Android 16, Chrome) the audio share fails in the PWA because of a reported lack of AAC, while the native Android app works fine sharing audio. If this isn't just a bug and there's no simple way to use a different method, it might be good to give a message to the user to the effect of "This feature requires the AAC audio codec to be installed on your phone", rather than silently failing.

image

TheNonPirate:
I tried adding another type of audio that it could try to fall back to. Does it work now?

benpaj-cps:
It gets the audio, but uses the download fallback. Also, the filename seems to only be set once per page load (i.e. all further downloads have the same filename, no matter the verse)?

benpaj-cps:
The ironic thing is that the Android app exports to `.m4a`, which is almost certainly using the AAC codec since I don't expect my Google phone to be using the "Apple Lossless Audio Codec". So either AAC is actually installed on my phone after all or the Android app is supplying it... I'll see if there's a simple way to check that.

TheNonPirate:
I fixed the filename only being set once per page load. I have found that the sharing option never seems to work, but I'm not sure if that's something I can fix or not.

benpaj-cps:
I think the AAC problem was due to my accessing it from my phone on LAN over HTTP, whereas AudioEncoder is only supported over HTTPS (see https://developer.mozilla.org/en-US/docs/Web/API/AudioEncoder). Running it on `localhost` and port forwarding 5173 from my laptop did work. It still gets the "download" behavior, but at least it's functional.

benpaj-cps:
The Android app adds a bit more information when sharing text.

PWA:
```
SAB Test PWA

The beginning of the Good News of Jesus Christ, the Son of God.
Mark 1:1
```

Android:
```
The beginning of the Good News of Jesus Christ, the Son of God. 
Mark 1:1
The World English Bible (WEB) is a Public Domain translation of the Holy Bible.
https://dwr8g.app.link?ref=ENGWEB/MRK.1.1
```

benpaj-cps:
However it looks like the URL link from the Android app doesn't work properly. Perhaps that will eventually be another config field, but that's likely a different issue.

TheNonPirate:
As far as I can tell, the copyright information (The World English Bible (WEB) is a Public Domain translation of the Holy Bible.) isn't even exported anywhere (It's not in config, for example). I think what the shared text says is a separate issue from sharing the audio.

benpaj-cps:
Right, okay. After a little digging I found that it comes from the `.appDef` but isn't converted by any of the current PWA scripts.

image

Since the share text functionality was in the new modal I didn't realize it predated this PR - that is probably a separate issue. I'll open one for completeness.

FyreByrd:
I get the following error when trying this on Mark 13, which has audio defined but no local files (i.e. the audio is downloaded from elsewhere). The audio itself plays fine in the PWA, it's only an issue with sharing.

`Access to fetch at 'https://dem-sab-assets-test.s3.amazonaws.com/web/B02___13_Mark________ENGWEBN2DA.webm' from origin 'http://localhost:5173' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

TheNonPirate:
I was able to figure out how to detect if it has a remote file and share it if it does (Although it seems to just share it as a link, not a file), but the closest I can find to a download option just opens a link to the file. Should I have that as the fallback in case sharing fails, or do something else? It looks like the usual method of fetching the file won't always work for remote files (If their server has the wrong permissions set for cross-site fetches).

chrisvire:
> I was able to figure out how to detect if it has a remote file and share it if it does (Although it seems to just share it as a link, not a file), but the closest I can find to a download option just opens a link to the file. Should I have that as the fallback in case sharing fails, or do something else? It looks like the usual method of fetching the file won't always work for remote files (If their server has the wrong permissions set for cross-site fetches).

If the timing files exist, they will be a local resource and will not require a cross-site fetch.

TheNonPirate:
In that case, what is going on with Mark 13? It seems to have remote audio, but it also seems to have timing files (It highlights verses as they are read in the audio).

chrisvire:
> In that case, what is going on with Mark 13? It seems to have remote audio, but it also seems to have timing files (It highlights verses as they are read in the audio).

It can have remote audio with local timing files.  The timing files will always be local.

Sharing a remote file really isn't an option. For FCBH Audio, it requires an API call to fetch the audio.

TheNonPirate:
What should I do about remote files, then? Should it just treat the chapter like one that doesn't have an audio file at all for the purposes of sharing?

chrisvire:
> What should I do about remote files, then? Should it just treat the chapter like one that doesn't have an audio file at all for the purposes of sharing?

On the Native app, the audio file has to be downloaded before it can be shared. The same thing should be true  for the PWA.

TheNonPirate:
How are we going to handle downloading audio files? They need to be downloaded somewhere where they can be fetched in the future, especially for offline use, right? Do we already have an issue for that? I was thinking I'd seen an issue about downloading audio, but now I can't find it.

FyreByrd:
> How are we going to handle downloading audio files? They need to be downloaded somewhere where they can be fetched in the future, especially for offline use, right? Do we already have an issue for that? I was thinking I'd seen an issue about downloading audio, but now I can't find it.

I couldn't find any associated issues either. I messaged @davidmoore1 to ask him to add the Issue(s) since he's the one who volunteered to document the feature. Should this PR move ahead without that feature? Or should we wait for it to be implemented? Thoughts @chrisvire?

FyreByrd:
I think this needs to wait on #1036 to be implemented. I am marking this as a draft PR until then.

FyreByrd:
> I get the following error when trying this on Mark 13, which has audio defined but no local files (i.e. the audio is downloaded from elsewhere). The audio itself plays fine in the PWA, it's only an issue with sharing.
> 
> `Access to fetch at 'https://dem-sab-assets-test.s3.amazonaws.com/web/B02___13_Mark________ENGWEBN2DA.webm' from origin 'http://localhost:5173' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

I have fixed this issue by modifying the CORS policy for the S3 bucket that contains the asset. I am also working on #1059 to add conversion-time checks that simulate the browser preflight check that would cause this issue at runtime in order to provide helpful warnings to users.

FyreByrd:
> How are we going to handle downloading audio files? They need to be downloaded somewhere where they can be fetched in the future, especially for offline use, right? Do we already have an issue for that? I was thinking I'd seen an issue about downloading audio, but now I can't find it.

I thought about it for a bit, and we probably don't actually need #1036 for this to work, especially now that I've fixed the CORS issue at the asset source. We may want to consider a user friendly error display...

src/lib/components/ShareSelector.svelte
```diff
@@ -0,0 +1,258 @@
...
+    let modalId = 'shareSelector';
+    let modalThis: Modal;
+    export function showModal(shareTextOnly: boolean) {
+        if (shareTextOnly) {
+            shareSelectedText();
+        } else {
+            modalThis.showModal();
+        }
+    }
+    const positioningCSS = $derived(
```
FyreByrd: Can you make this a bit more dry in its usage across here, TextAppearanceSelector, DownloadSelector, and PlanStopDialog?

src/lib/components/ShareSelector.svelte
```diff
@@ -0,0 +1,258 @@
...
+            {$t['Share_Text']}
+        </button>
+        <button
+            class="dy-btn dy-btn-sm flex items-center justify-center gap-2"
+            onclick={() => shareAudio()}
+        >
+            <AudioIcon.Volume />
+            {$t['Share_Audio']}
+        </button>
+        <!--<button
```
FyreByrd, line 198: Why is this commented out?

src/lib/components/TextSelectionToolbar.svelte
```diff
@@ -7,7 +7,6 @@ TODO:
  -> Share
  -> Play
  -> Play Repeat
- -> Verse On Image
```
FyreByrd, line 10: I think all of these TODOs can be removed?

src/lib/components/ShareSelector.svelte
FyreByrd, line 1: I think this is too much of a discrepancy...

Image

Image

FyreByrd (changes requested):

FyreByrd:
Ack! I somehow forgot to actually test it...

For some reason it always fails the share attempt on WEB then goes to download...
But on CUK it succeeds in the share call, but because it succeeds I am never able to directly download it...

(of course the behavior of navigator.share is platform specific, so idk how much success you'll have reproducing...)

Also, I would recommend making the share portion of shareAudio in ShareSelector be more DRY with share.ts

src/lib/components/ShareSelector.svelte
TheNonPirate: It's commented out because video download has not yet been implemented, but that is where the button would go if it was implemented.

TheNonPirate:
I think that the reason is fails the share attempt on WEB is that the files you were trying to share were probably local files. It seems that the file types that I had it prioritize exporting are ones that Brave does not allow sharing for. I am also finding that the share UI doesn't seem to be very helpful for actually sharing the file, at least on Brave. I have not managed to share the file yet in my testing. Do you have any suggestions for something better than relying on navigator.share?

src/lib/components/ShareSelector.svelte
TheNonPirate: As I've looked through those 4 modals, I've found that PlanStopDialog didn't actually use its positioningCSS and each of the others had slightly different positioningCSS because each of them belonged in a different position. What is a good way to make it more DRY given that? Should I just make a common function for the `(Number(vertOffset.replace('rem', '')) + 1) + 'rem;'` calculation?

FyreByrd: If the difference is just, say, `top` or `bottom`, you can make that a parameter of the function

TheNonPirate: Two of them use top, and one uses bottom. Two of them use `width:auto;`, and one doesn't.

FyreByrd: Gotcha, I feel like `width:auto` could be treated separately... it has nothing to do with positioning

FyreByrd:
> I think that the reason is fails the share attempt on WEB is that the files you were trying to share were probably local files. It seems that the file types that I had it prioritize exporting are ones that Brave does not allow sharing for. I am also finding that the share UI doesn't seem to be very helpful for actually sharing the file, at least on Brave. I have not managed to share the file yet in my testing. Do you have any suggestions for something better than relying on navigator.share?

I'm not sure why local files make a difference??? I'm not sure if there's something better than navigator.share or not... May be a question for @davidmoore1 ?

TheNonPirate:
Local files make a difference because (At least when it fails to fetch) remote files share a link rather than a file, so it's just text, which is much easier to share.

FyreByrd (approved):
With a couple tweaks made in my commit, this functions quite well. Please review the changes that I made before I merge.

FyreByrd:
Also, the issue I ran into much earlier is because it was trying to share with an invalid MIME type (i.e. `audio/mp4`). I remedied this be using the `mimeType` instance variable from the format object.

FyreByrd:
@coderabbitai review

PR #996, opened 2026-06-15
#995 Fix CSS custom attribute declaration for border image source

### LINKED ISSUE #995: Border image doesn't render
With the PWA test app data, borders are enabled for Psalms but don't show in the PWA. The blank area around the screen edge is there, but there is no image.

PWA on left, Android on right:

Image

Image

benpaj-cps:
Fixes #995. When the border image source was parameterized, the CSS variable wasn't declared [properly](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/--*), so no border images were being rendered.

Before/after/Android:

image

image

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed border image styling to display correctly.

### DIFF
--- a/src/routes/text/+page.svelte
+++ b/src/routes/text/+page.svelte
@@ -564,7 +564,7 @@
             {/if}
         {/if}
         <div
-            style="--borderImageSource=url({borders['./border.png']});"
+            style="--borderImageSource: url({borders['./border.png']});"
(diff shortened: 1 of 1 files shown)


FyreByrd (approved):
Good catch!

PR #1027, opened 2026-06-24
Tweak new collections route

### LINKED ISSUE #1026: Collection Icons Hardcoded to Black
The icons such as SinglePaneIcon used for collection selection are currently hardcoded to black. In the native app, they appear white in dark mode. However, it is unclear what does determine their color. It seems that the icons cannot be inspected using CSS debugging, and even when I tried changing everything in both Main Colors and Book Colors to extremely bright colors (Most red, blue, and green) for the Sepia theme, the icon stays black in the Sepia theme.
Image
Aidan has done as thorough of an investigation as he can of the app-builders codebase, and he has no idea where the icon color is coming from. He found one function in ColorStringUtils.java that looked promising, but it does not appear to be used anywhere.

According to Chris, we just need to use white for the dark theme and white for the others.

chrisvire:
- layout is a better name for the route
- some of the tailwind layout classes should have been on the container not on the component
- when it is ONLY the Single Pane, the Native app doesn't display the title.
- when it is ONLY another option other than Single Pane, the Native app does display the title

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated layout selection screen and updated related navigation to open it directly.
  * Layout option menu titles are now shown conditionally based on enabled layout choices.

* **Bug Fixes**
  * Corrected multiple navigation targets (sidebar, content links, and text page redirects/buttons) to send users to the layout page instead of the collections page.
  * Improved the layout page’s tab container to better control scrolling and spacing.

### DIFF
--- a/src/routes/layout/+page.svelte
+++ b/src/routes/layout/+page.svelte
@@ -12,7 +12,8 @@
         monoIconColor,
         refs,
         selectedLayouts,
-        t
+        t,
+        testLayouts
     } from '$lib/data/stores';
     import { CheckIcon, SideBySideIcon, SinglePaneIcon, VerseByVerseIcon } from '$lib/icons';
 
@@ -40,9 +41,11 @@
     // ToDo: If showSinglePane false, provide first availible visible option instead
     const showSinglePane = !!scriptureConfig.layouts?.find((x) => x.mode === Layout.Single)
         ?.enabled;
-    const showSideBySide = !!scriptureConfig.layouts?.find((x) => x.mode === Layout.Two)?.enabled; //Not yet implemented
-    const showVerseByVerse = !!scriptureConfig.layouts?.find((x) => x.mode === Layout.VerseByVerse)
-        ?.enabled; //Not yet implemented
+    const showSideBySide =
+        testLayouts || !!scriptureConfig.layouts?.find((x) => x.mode === Layout.Two)?.enabled; //Not yet implemented
+    const showVerseByVerse =
+        testLayouts ||
+        !!scriptureConfig.layouts?.find((x) => x.mode === Layout.VerseByVerse)?.enabled; //Not yet implemented
 
     // In the native app, if only showing the single pane, then don't show the title.
     // If showing one of the other two, then show the title.
--- a/src/routes/layout/+page.svelte
+++ b/src/routes/layout/+page.svelte
@@ -35,7 +35,10 @@
     const showSideBySide = !!scriptureConfig.layouts?.find((x) => x.mode === Layout.Two)?.enabled; //Not yet implemented
     const showVerseByVerse = !!scriptureConfig.layouts?.find((x) => x.mode === Layout.VerseByVerse)
         ?.enabled; //Not yet implemented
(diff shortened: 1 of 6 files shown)


src/routes/layout/+page.svelte
```diff
@@ -35,6 +35,7 @@
     const showSideBySide = !!scriptureConfig.layouts?.find((x) => x.mode === Layout.Two)?.enabled; //Not yet implemented
     const showVerseByVerse = !!scriptureConfig.layouts?.find((x) => x.mode === Layout.VerseByVerse)
         ?.enabled; //Not yet implemented
+    const multipleLayouts = showSideBySide || showVerseByVerse;
```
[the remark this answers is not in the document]
chrisvire: No, the title is not shown if it is just layout single. I should add a comment.

FyreByrd (approved):

TheNonPirate:
Should there be a scrollbar here? It seems like it should take up the height of the page instead of being cut short and using a scrollbar.

image

chrisvire:
> Should there be a scrollbar here? It seems like it should take up the height of the page instead of being cut short and using a scrollbar.

@TheNonPirate What does the native app do with this project?

TheNonPirate:
It lets the collections go to the bottom of the screen (I didn't use a good screenshot; I should have made it more clear in my screenshot above that I didn't screenshot the entire screen; the scrollbar only went down about half the width of the screen).
Here's a better screenshot of the PWA:
image

src/routes/layout/+page.svelte
FyreByrd, line 1: There's a bit of an issue with scrolling... 

Image

Image

Based on a preliminary look, you will need to specify the height property on TabsMenu (default value is 50vh, i.e. half the screen). I think you may also need to set height/scroll around LayoutOptions, so that the scroll behavior doesn't cause the tab icons to disappear.

FyreByrd: If you use `$monoIconColor` from `$lib/data/stores/theme` here (and with the other icons), you could close #1026. I think this PR would be a good place to do so...

FyreByrd (changes requested):

src/routes/layout/+page.svelte
chrisvire: Yep, I can get that here.

FyreByrd: This hasn't been fixed yet. Everything else looks fine.

FyreByrd (changes requested):

FyreByrd:
> Should there be a scrollbar here? It seems like it should take up the height of the page instead of being cut short and using a scrollbar.
> 
> image

This is the same issue as I am referring to. It's mostly because TabsMenu has a default max-height of 50% of the screen, which is not what we want here. I think an additional tweak would be needed on top of fixing the max-height in order to prevent the tabs from being able to be scrolled out of view in some cases (probably only relevant on Single Pane with a large number of collections)

FyreByrd (approved):
I think this is fine as is. Trying to fix the scrolling buttons further is more trouble than it's really worth.

PR #1031, opened 2026-06-24
Redirect to First Book In Collection if Current Book is Undefined

### LINKED ISSUE #1028: Changing to a collection that doesn't contain the current item fails to redirect
When switching to a collection that does not contain the chapter, song, or story currently being viewed, the interface is stuck with an endless spinner (and an empty border, if that is configured).

In the Android app, changing to a collection that does not contain the current item redirects to the first available item.

https://github.com/user-attachments/assets/d8b4770b-c417-4c08-86e6-4e6426847eb2

TheNonPirate:
Fixes #1028. The text route now checks if the current book is undefined, and if it is, it redirects to the first book in the collection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved book selection handling on the text page so it can recover from stale or invalid selections.
  * If the chosen book is no longer available in the selected collection, the page now automatically falls back to the first valid book.

### DIFF
--- a/src/routes/text/+page.svelte
+++ b/src/routes/text/+page.svelte
@@ -88,6 +88,21 @@
     }
 
     let { data }: Props = $props();
+    $effect(() => {
+        const book = scriptureConfig?.bookCollections
+            ?.find((x) => x.id === $refs.collection)
+            ?.books.find((x) => x.id === $refs.book);
+        if (!book) {
+            console.log('No book!');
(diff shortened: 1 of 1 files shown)


FyreByrd:
I've already been working on something similar in #1029, but handling the issue on the reference parsing level.

chrisvire:
Will be fixed in #1029

PR #1032, opened 2026-06-26
Implement Quiz Access Codes

### LINKED ISSUE #551: Quiz may be restricted: Code entry required
A quiz may require the entry of a code before it can be unlocked.  After it is unlocked, it remains unlocked and may be run as often as desired without seeing the code entry screen again (so a requirement is that you save the locked unlocked status in storage).  When locked, if the quiz book is selected, a screen is displayed with the name of the quiz as the title and then a message (Quiz_Access_Code_Message) is displayed saying "Please enter the access code to unlock this quiz".  Finally a numeric keyboard is displayed at the bottom.   An unlock icon serves as the "Check code" button.  If this button is pressed and the code matches the configured code, then the first page of the quiz is displayed.  If the code is wrong, nothing happens.  There isn't an error message or any error indicator.  It just doesn't do anything.

The quiz is defined as a book in configuration.  The book configuration contains a quiz section which contains a features list.   If the feature "access-type" is defined and has a value of "code" then the code entry is required.  The code to be matched is in feature "access-code".  An example of the book configuration for a quiz with this lock is: 
   <book id="QU2" type="quiz">
      <name>QU2</name>

      <font-choice type="book-collection" />
      <quiz>
        <features type="quiz">
          <feature name="shuffle-questions" value="true"/>
          <feature name="auto-advance" value="false"/>
          <feature name="auto-advance-delay" value="2000"/>
          <feature name="access-type" value="code"/>
          <feature name="access-code" value="1234"/>
        </features>
      </quiz>
      <filename>quizbug.txt</filename>
      <source>/Users/david/App Builder/Scripture Apps/quizbug.txt</source>
      <features type="book">
        <feature name="shuffle-questions" value="true"/>
        <feature name="auto-advance" value="false"/>
        <feature name="auto-advance-delay" value="2000"/>
        <feature name="access-type" value="after"/>
        <feature name="access-after" value="QUIZ1"/>
      </features>
    </book>

An example of the screen displayed is: 
EnterQuizCode

TheNonPirate:
Resolve #551. If a quiz is set to require a password to enter, then when the quiz is navigated to, it will display a screen with a numpad to input the password. Once it is input, it does not need to be input again due to a database of unlocked quizzes.

## Summary by CodeRabbit

* **New Features**
  * Added code-based quiz access with a numeric keypad, backspace, and unlock control.
  * Persist unlock status locally so already-unlocked quizzes remain accessible on revisit.
  * Added backspace and open-lock icons for the unlock interface.
* **Bug Fixes**
  * Improved the locked quiz screen to correctly handle dependent quizzes vs access-code prompts.
  * Ensured configured access codes preserve leading zeros.

### DIFF
--- a/src/routes/quiz/[collection]/[id]/+page.svelte
+++ b/src/routes/quiz/[collection]/[id]/+page.svelte
@@ -259,10 +259,7 @@
     }
     function attemptUnlock() {
         if (accessCode?.toString() === passwordInput) {
-            if (book?.quizFeatures) {
-                locked = false;
-                addQuizUnlocked({ collection, book: quizId }).then(() => invalidateAll());
-            }
+            addQuizUnlocked({ collection, book: quizId }).then(() => invalidateAll());
         }
     }
     function handleKeyInput(e: KeyboardEvent) {
@@ -317,7 +314,7 @@
     {#if locked}
         {#if data.dependentQuizName || data.dependentQuizId}
             <div class="quiz-locked">
-                <div class="quiz-locked-title">{quizId}</div>
+                <div class="quiz-locked-title">{displayLabel}</div>
                 <div class="quiz-locked-message">
                     {$t['Quiz_Access_After_Message']}
                 </div>
@@ -325,7 +322,7 @@
             </div>
         {:else if accessCode}
             <div class="quiz-locked">
-                <div class="quiz-locked-title">{quizId}</div>
+                <div class="quiz-locked-title">{displayLabel}</div>
                 <div class="quiz-locked-message">
                     {$t['Quiz_Access_Code_Message']}
                 </div>
--- a/src/routes/quiz/[collection]/[id]/+page.svelte
+++ b/src/routes/quiz/[collection]/[id]/+page.svelte
@@ -321,9 +321,9 @@
                 <div class="quiz-locked-name">{data.dependentQuizName || data.dependentQuizId}</div>
             </div>
         {:else if accessCode}
-            <div class="quiz-locked">
-                <div class="quiz-locked-title">{displayLabel}</div>
-                <div class="quiz-locked-message">
+            <div class="quiz-keypad max-w-breakpoint-md mx-auto">
+                <div class="quiz-keypad-title">{displayLabel}</div>
+                <div class="quiz-keypad-message">
                     {$t['Quiz_Access_Code_Message']}
                 </div>
                 <input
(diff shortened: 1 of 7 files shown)


FyreByrd (changes requested):
How did you test this? I'd appreciate having a testable quiz added to the project data...

TheNonPirate:
I updated the sab_pwa_test quiz, fixing a problem with its illustrations (They didn't all show up in either the PWA or the native app) and giving it an access code.

src/routes/quiz/[collection]/[id]/+page.svelte
```diff
@@ -281,13 +312,65 @@
...
+                    bind:value={passwordInput}
+                    readonly
+                    onkeydown={handleKeyInput}
+                />
+            </div>
+            <div class="grid grid-cols-3 gap-4 text-center p-10 text-[30px]">
+                {#each Array(9) as _, i}
+                    <!-- svelte-ignore a11y_click_events_have_key_events -->
+                    <!-- svelte-ignore a11y_no_static_element_interactions -->
+                    <div
```
Comment (FyreByrd) [inline src/routes/quiz/[collection]/[id]/+page.svelte]: Keypad inputs should be button instead of div. Please also extract as many of these shared styles as you can into a `<style>` element instead.

FyreByrd (changes requested):

src/routes/quiz/[collection]/[id]/+page.svelte
Comment (TheNonPirate) [reply src/routes/quiz/[collection]/[id]/+page.svelte]: As in, a CSS class in a style element in the same script?

src/routes/quiz/[collection]/[id]/+page.svelte
```diff
@@ -281,13 +312,65 @@
...
+                <div class="quiz-locked-message">
+                    {$t['Quiz_Access_After_Message']}
+                </div>
+                <div class="quiz-locked-name">{data.dependentQuizName || data.dependentQuizId}</div>
             </div>
-            <div class="quiz-locked-name">{data.dependentQuizName || data.dependentQuizId}</div>
-        </div>
+        {:else if accessCode}
+            <div class="quiz-locked">
+                <div class="quiz-locked-title">{displayLabel}</div>
```
Comment (FyreByrd) [inline src/routes/quiz/[collection]/[id]/+page.svelte]: This should be `quiz-keypad-title`. Similar comment for below.

FyreByrd:
Don't know how I completely forgot this, but we should have max width on the quiz as well...
Screenshot 2026-07-01 at 5 03 10 PM

test_data/projects/sab/sab_pwa_test.zip
chrisvire, line 1: Let's have a separate PR with updates to the sab_pwa_test project

FyreByrd:
This needs to be rebased on main for the updated project data in #1045

FyreByrd (approved):

TheNonPirate:
Are the lint/test errors an issue, or is that just because the new color tests are causing it to fail?

FyreByrd:
CI is failing due to changes made in SAB 14.3, which is still in the process of being released today. As the most recent change was just rebasing on main to bring in updated project data, I think this should be fine as is.

Issue #1000, opened 2026-06-16
Dark Mode: Note icon and editing


chrisvire:
There are issues with styling in dark mode with notes:

1. the note icon is black in dark mode (check native app to see which style is used for note icon)
2. when clicking on the note icon in dark mode, the popup text is not readable
3. when editing a note in dark mode, the editing window should not have black text on a white background

There is a note on verse 1:

<img width="612" height="1120" alt="Image" src="https://github.com/user-attachments/assets/c9580899-f0af-4744-bff9-0f7950df52a4" />

<img width="612" height="1120" alt="Image" src="https://github.com/user-attachments/assets/d8c1e6eb-a636-4e38-b2c4-e75002a21ed9" />

Clicking on the note:

<img width="612" height="1120" alt="Image" src="https://github.com/user-attachments/assets/f1c761cf-a195-4ec8-af64-1e764592edea" />

<img width="612" height="1120" alt="Image" src="https://github.com/user-attachments/assets/b51412bb-cf2b-4b37-8a50-c4a26f14bec1" />

Editing the note (clicking on the pencil):

<img width="612" height="1120" alt="Image" src="https://github.com/user-attachments/assets/0a862154-8721-48bb-8ed4-fe8eac34397d" />

<img width="612" height="1120" alt="Image" src="https://github.com/user-attachments/assets/f9d58a1e-70ed-490b-a0ce-d6d840673326" />

TheNonPirate:
The proper styling is in src/gen-assets/styles/sab-app.css under selectors such as `#container.annotation-note`.

benpaj-cps:
This is one of several instances where the app currently uses a default DaisyUI style without overriding it. One problem is that DaisyUI is always set to the `light` theme regardless of user theme choice. I'm having to move away from several of the other DaisyUI defaults in the v5 upgrade to keep the styles consistent with the current PWA.

benpaj-cps:
The note icon is black in dark mode because it is a hardcoded SVG and not the actual `NoteIcon`.

<img width="2254" height="248" alt="Image" src="https://github.com/user-attachments/assets/ccc48ea9-9047-4110-a388-8931da8104f5" />

chrisvire:
@benpaj-cps, try this:

```
    const noteSvg = () => {
        let textColor = themeColors?.TextColor || '#000000';
[...]

PR #1034, opened 2026-06-29
Fix dark/sepia mode color for inline note icon

benpaj-cps:
Changed in-text bookmark icon to use current text color. Falls back to `monoIconColor` if the theme's text color is not defined.

Before:
image

After:
image
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Note icons now better match the app’s theme colors, with a fallback that improves visibility when theme values aren’t available.
  * Updated icon styling to avoid the previous hardcoded black color, resulting in more consistent appearance across themes.

### DIFF
--- a/src/lib/components/ScriptureViewSofria.svelte
+++ b/src/lib/components/ScriptureViewSofria.svelte
@@ -1,6 +1,6 @@
 <!--
 @component
-A component for displaying scripture.  
+A component for displaying scripture.
 TODO:
 - find a way to scroll smoothly, as CSS only option does not work as expected.
 - save graft info so that references can be handled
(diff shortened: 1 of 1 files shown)


FyreByrd (changes requested):
#1000 also references issues with the note edit page. Those would need to be fixed before this could be considered as closing that issue.

benpaj-cps:
> #1000 also references issues with the note edit page. Those would need to be fixed before this could be considered as closing that issue.

Right. That will probably come in a separate PR as discussed there. Would you like me to unlink this PR so it can be dealt with on its own, and have the styling PR close #1000?

FyreByrd:
> Right. That will probably come in a separate PR as discussed there. Would you like me to unlink this PR so it can be dealt with on its own, and have the styling PR close #1000?

Yes please. You can still reference #1000, just unlink it.

FyreByrd (approved):
Approved pending #1000 being unlinked.

PR #1035, opened 2026-06-29
Give Modal Component Variable Styles

### LINKED ISSUE #1018: Modal Component has Hardcoded Styles
Right now, the Modal component uses this code for its styling: `style={convertStyle($s?.['ui.dialog']) + addCSS}`
This allows CSS to be added but always uses the ui.dialog style instead of using the styling that is needed for that particular component.  If nothing else, the TextAppearanceSelector has the wrong background color currently.

TheNonPirate:
Closes #1018. The Modal component now has a property that allows its styling to be specified instead of being hardcoded, and each component that used the Modal now has its styling matching that of the native app. They were also moved inside the container div in +layout.svelte so they could use the CSS styling variables.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated dialogs and modals to use consistent, theme-based background styling for improved visual uniformity.
  * Refined text appearance controls so numeric font-size and line-height labels display more clearly.
* **Refactor**
  * Reorganized where popup dialogs render within the sidebar for a cleaner, more reliable layout.

### DIFF
--- a/src/lib/components/CollectionModal.svelte
+++ b/src/lib/components/CollectionModal.svelte
@@ -96,7 +96,7 @@
     function handleOk() {}
 </script>
 
-<Modal bind:this={modal} {id}>
+<Modal bind:this={modal} {id} styling={'background-color:var(--BackgroundColor);'}>
     <div class="text-center">
         <div class="overflow-y-auto max-h-[80vh] p-1 text-left">
             <CollectionList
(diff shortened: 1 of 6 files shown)


FyreByrd (changes requested):
Looks pretty good to me. Only thing I would change is to use the `$themes` store instead of `var(--RandomColor)` (sorry, this is partially on me for not fixing some of the other usage sooner)

TheNonPirate:
Where is the `$themes` store? I can't find any store called `$themes`. There is a themes variable exported from stores, but it's just an array of strings containing the 3 names of the themes.

FyreByrd:
> Where is the `$themes` store? I can't find any store called `$themes`. There is a themes variable exported from stores, but it's just an array of strings containing the 3 names of the themes.

`$themeColors` sorry

FyreByrd (approved):

PR #998, opened 2026-06-15
Improve types for annotations and scripture view

FyreByrd:
I recommend rebase and merge, since these commits are not super closely related.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced scripture view reliability with corrected plan section visibility and more dependable in-text navigation (including footnote/link handling and media placement)
  * More robust verse/phrase selection syncing to the current selection state, with safer click targeting
  * Stronger sorting consistency across bookmarks, highlights, and notes using a unified sort mode
* **Bug Fixes**
  * Improved image cropping stability with added guards for missing images and null-safe crop results
* **Maintenance**
  * Improved type safety across the affected UI features and stores for greater runtime confidence

PR #1038, opened 2026-06-29
Implement verse by verse layout

### LINKED ISSUE #593: Verse by Verse Display
Verse by verse mode allows the user to select to view multiple book collections at the same time in a single pane.  For each verse, the verse is displayed for the first book collection, then the second, and, if setup for three, the third.  In SAB, this is configured in the Books section, at the top level, under the Layouts tab.  The user can select whether verse by verse is available and, if it is, what the default book selections are for that mode.

This appears in the config.js file under Layouts: (I believe convertConfig is forcing enable to be false and needs to be changed).  As far as I can see, the features is only used with the two pane view so it should always be empty for verse by verse.
```
    {
      "mode": "verse-by-verse",
      "enabled": false,
      "layoutCollections": [
        "CUKNVS",
        "ENGWEB"
      ],
      "features": {}
    }
```

When more than one layout mode is configured, the layout screen has an additional tab selection bar added to it.  When the layout screen displays, it comes up with the tab for the currently selected mode active.  The following example shows for all three modes selected, but if just single mode and verse by verse are configured, then the middle tab would not be present.
SinglePane

When the verse by verse tab is selected, then the layout screen that allows up to three book collection selections appears.  This screen will always have at least two entries.   The third collection dropdown should only be displayed if more than 2 book collections are configured in this app.  Otherwise you would only see the first two dropdowns.  The third dropdown may be empty, so all dashes is an option in the display window along with the remaining book collections.  In each case, if the user selects a book collection that is already selected by one of the other entries, the other entry with that book collection is changed to the first unselected book collection.

It probably should be mentioned here how restricted book collections are handled.  Issue #567 references that book collections may be restricted so that they may only appear along with other book collections.  While the single pane view will filter out restricted book collections, all book collections should appear in all three slots in verse by verse mode.

VerseByVerseSelect

When the verse is displayed, the verse number is displayed, then the text for the first book collection, followed by the text for the second followed by the text for the third.  
VerseByVerseDisplay

Note that only the text for the first book collection is selectable (for bookmarks highlighting and notes).  The markers for notes and bookmarks appear after the text for the first book collection.
verseByVerseAnnotations

Here is the HTML as it appears in the Android app.  Note that the verse overall appears in a new block div with classes that appear to alternate between block1 and block2.  The div with the primary book collection text is the same as what appears in the single pane class, so it should just be a case of reproducing what is there. If it is broken into sections for audio, then it still appears just as it did in the original single pane This is followed by a verse-by-verse-block class which contains a div with classes verse-by-verse-verse and verse-by-verse-1 with the text of the 2nd book collection. If a third collection is specified, then it is in a div with verse-by-verse-verse and verse-by-verse-2 class.  These are not subdivided into sections since there is no audio associated with them.  
Screenshot 2024-06-07 at 9 52 42 AM

TheNonPirate:
Resolves #593. This PR implements the verse-by-verse display. This was done by making the Proskomma renderer run once for each collection, but for all but the first, it adds components to the HTML that has already been rendered instead of rendering it again from scratch.

## Summary by CodeRabbit

* **New Features**
  * Remember layout selection and the active layout mode across sessions.
  * In verse-by-verse mode, the view can load additional document sets for the selected layouts.

* **Bug Fixes**
  * Improved verse-by-verse rendering consistency across multi-pass rendering.
  * Prevented duplicated top-level content and repeated UI setup during subsequent render passes.
  * Corrected layout availability so verse-by-verse layouts are no longer incorrectly disabled.

### DIFF
--- a/convert/convertConfig.ts
+++ b/convert/convertConfig.ts
@@ -1606,10 +1606,10 @@ function filterFeaturesNotReady(data: ScriptureConfig | DictionaryConfig) {
 
     // Two pane and Verse-By-Verse are not done
     if (isScriptureConfig(data)) {
-        // Two pane and Verse-By-Verse are not done
+        // Two pane is not done
         if (data.layouts) {
             for (const layout of data.layouts) {
-                if (layout.mode === 'two' || layout.mode === 'verse-by-verse') {
+                if (layout.mode === 'two') {
                     layout.enabled = false;
                 }
             }
--- a/src/lib/components/ScriptureViewSofria.svelte
+++ b/src/lib/components/ScriptureViewSofria.svelte
@@ -61,7 +61,6 @@ LOGGING:
         monoIconColor,
         plan,
         refs,
-        selectedLayouts,
         t,
         userSettings
     } from '$lib/data/stores';
@@ -1924,10 +1923,12 @@ LOGGING:
                                 }
 
                                 if (verseByVerseMode) {
-                                    workspace.paragraphDiv.appendChild(
-                                        workspace.verseByVerseDivRoot.cloneNode(true)
-                                    );
-                                    workspace.verseByVerseDivRoot = null;
+                                    if (workspace.verseByVerseDivRoot) {
+                                        workspace.paragraphDiv.appendChild(
+                                            workspace.verseByVerseDivRoot.cloneNode(true)
+                                        );
+                                        workspace.verseByVerseDivRoot = null;
+                                    }
                                 }
 
                                 workspace.phraseDiv = null;
--- a/src/lib/components/ScriptureViewSofria.svelte
+++ b/src/lib/components/ScriptureViewSofria.svelte
@@ -321,7 +321,10 @@ LOGGING:
                     break;
             }
             const phraseIndex = createLetterIndex(workspace.currentPhraseIndex);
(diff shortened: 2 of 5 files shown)


FyreByrd (changes requested):
This is an impressive first start. A few things stick out that need to be changed. See the screenshots below.

1. The font is set only by the first collection (contrary to native).
2. The LTR/RTL mix is all wrong (sorry... you don't deserve to deal with this...).
3. The layout should be persisted in localStorage (i.e. page refresh currently resets to single pane).
4. Related to above, native app shows current layout when going to layout page, whereas PWA always shows single pane first.

(WEB, ASV, MOR)
Image

(MOR, ASV, WEB)
Image

FyreByrd:
Also, I have some updated type changes in #998. I don't know if you would want these or not for this PR.

TheNonPirate:
New problems have been identified:
The book header on chapter 1 of a book appears at the bottom in each selected translation except the first one.
Mark 1 is really messed up in any combination of collections that includes WEB.

TheNonPirate:
Mark 1 is messed up if WEB is the first collection, it seems, not if it's just one of the collections.

FyreByrd:
Marking as draft, pending rendering refactor by @benpaj-cps to support #1039

PR #1039, opened 2026-06-30
#459 Render rich text in cross-ref pop-up

### LINKED ISSUE #459: Improve display of reference popups
Currently when references are displayed in a pop up, the text for the referenced text is displayed as raw text, without verse numbers or embedded references.  The iOS and Android versions display the text much as it displays in the main window.  Embedded references can be selected from the reference popup, causing a stack of popups if the user doesn't select to move to that entry.

benpaj-cps:
In progress, closes #459.

- Fix parent div capturing all clicks on footnotes (separated as #1058).
- ...

### DIFF
--- a/src/lib/components/ScriptureViewSofria.svelte
+++ b/src/lib/components/ScriptureViewSofria.svelte
@@ -1428,6 +1428,8 @@ LOGGING:
     }
     // handles on click when interacting with the scripture view
     function onClick(e: any) {
+        const t = e.target as HTMLElement;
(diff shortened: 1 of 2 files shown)


benpaj-cps:
Rich text rendering in footnotes is pending figuring out how to get `ScriptureViewSofria` to efficiently render small verse ranges.

benpaj-cps:
The main issue in the PWA is that the native app renders almost all text features (headings, references, etc.), including user data such as bookmarks and notes, while the PWA only renders the bare text without the usual verse numbers. Screenshots to follow.

PR #1040, opened 2026-06-30
Change manifest to prevent icon cropping on Android

### LINKED ISSUE #1013: Improve icon on Android
We use the app icon from the Android app. It looks fine there, but when it is used to add the app to the homescreen, it look cropped.

Image

You can see the PWA (circled in red) and the native app icons.

Image

TheNonPirate:
Fixes #1013. The manifest is now changed to give icons the "purpose": "any" attribute, which fixes the icon cropping issue on Android.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved manifest output normalization for icon metadata, updating `purpose` values so supported icons are handled more consistently.
  * Adjusted icon path rewriting so generated manifest entries continue to point to the correct final icon files.

### DIFF
--- a/convert/convertManifest.ts
+++ b/convert/convertManifest.ts
@@ -54,8 +54,10 @@ export function convertManifest(dataDir: string, verbose: number) {
                     } else {
                         throw new Error(`Required icon file ${iconPath} does not exist!`);
                     }
-
                     line = line.replace(srcMatch[0], `"src": "${joinUrlPath('.', finalName)}"`);
+                }
+                if (line.includes('purpose')) {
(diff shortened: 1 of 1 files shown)


chrisvire:
@TheNonPirate if this is a bug in the manifest.json generation, then we should fix it in SAB.

chrisvire:
The SAB code hardcodes the generation of the manifest.json file.  So I changed generating the adaptive icon to use "any" instead of "maskable" and it has the same effect. So we don't need this change. The fix will be in 14.3 which should be release today or this week.

PR #1050, opened 2026-07-06
Save selectedVerses to localStorage

### LINKED ISSUE #1033: Note editor loses reference on page refresh but stays open
When adding a note on a verse,
Image

...and refreshing the page, and clicking the check button to save...
Image

...both the reference and the note text are lost, leading to an error, and the note cannot be recorded.
Image

I don't believe this can occur on the Android app, since the webview can't be as easily refreshed.

Two ways to deal with this might be:
  (1) always exit the note editor on page refresh (which might be annoying if drafted text is lost), or
  (2) store the current note text and reference locally, so a user can resume editing after a refresh.

TheNonPirate:
Fixes #1033. Back up $selectedVerses to localStorage so the note editor doesn't lose its reference on page reference.

## Summary by CodeRabbit

* **New Features**
  * Selected verses are now saved in the browser, so your current selection is preserved after refreshing or reopening the app.

### DIFF
--- a/src/lib/data/stores/scripture.ts
+++ b/src/lib/data/stores/scripture.ts
@@ -207,10 +207,12 @@ export type Selection = {
     reference: string;
     verse: string;
 };
-
+setDefaultStorage('selectedVerses', JSON.stringify([]));
 function createSelectedVerses() {
-    const external: Writable<Selection[]> = writable([]);
-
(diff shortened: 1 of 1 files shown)


FyreByrd (approved):

chrisvire:
Rebased off of main to pickup fixes for tests.

PR #1056, opened 2026-07-07
Fix for missing bar under ui.bottom-navigation

eomerdws:
Missing "bar" from ui.bottom-navigation caused the PWA to crash on launch with error about not being able to find 'background-color' from undefined.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the bottom navigation bar background color to correctly read the intended UI style setting, restoring the intended appearance.

### DIFF
--- a/src/lib/components/BottomNavigationBar.svelte
+++ b/src/lib/components/BottomNavigationBar.svelte
@@ -19,7 +19,7 @@
 
     const bottomNavBarItems = config?.bottomNavBarItems;
 
-    const barBackgroundColor = $derived($s?.['ui.bottom-navigation.']['background-color']);
+    const barBackgroundColor = $derived($s?.['ui.bottom-navigation.bar']['background-color']);
     const barTextColor = $derived($s?.['ui.bottom-navigation.item.text']['color']);
     const barTextSelectedColor = $derived($s?.['ui.bottom-navigation.item.text.selected']['color']);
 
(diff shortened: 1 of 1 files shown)


FyreByrd (approved):

src/lib/components/BottomNavigationBar.svelte
```diff
@@ -19,7 +19,7 @@
 
     const bottomNavBarItems = config?.bottomNavBarItems;
 
-    const barBackgroundColor = $derived($s?.['ui.bottom-navigation.']['background-color']);
```
chrisvire, line 22: For backward compatibility, should you check both for a little bit?

chrisvire: Sorry about changing that without seeing what would break

eomerdws: I hadn't realized this was due to a previous version. I had assumed it was a typo. Is there another way we want to handle it?

FyreByrd: How many SAB versions are affected by this?

FyreByrd: Try:

```js
($s?.['ui.bottom-navigation.bar'] ?? $s?.['ui.bottom-navigation.'])?.['background-color']
```

PR #1069, opened 2026-07-09
Implement audio download

### LINKED ISSUE #1036: Download audio for offline use
- A Scripture App Builder project can have multiple audio sources which allow either streaming audio, downloading audio, or both.  The PWA only supports streaming at the present time. This would allow apps that have been "Add to Homescreen" to store audio files in file storage to allow offline use.
- Implement a download manager, like the native app, to queue the downloading of audio files per book and per chapter. This allows a user who doesn't have consistent internet access to download all at once while they have access.

@davidmoore1 the project document has you listed for filling out further details. I've added what is in the document for now so we can start tracking the issue.

TheNonPirate:
Resolves #1036. This PR implements audio downloading using an internal database to store audio.

## Summary by CodeRabbit

* **New Features**
  * Added offline-capable audio downloads with persistent storage, including an MRU-style URL cache for faster reuse.
  * Introduced a download modal with progress tracking, cancellation, and transient error messaging.
  * Audio playback now checks availability before starting and prompts or auto-triggers downloads based on settings.
* **Bug Fixes**
  * Improved playback reliability when switching players and when loading metadata.
  * Hardened audio UI rendering when availability refs are not yet ready.
* **Style**
  * Refined modal styling default behavior and formatting.

### DIFF
--- a/src/lib/components/AudioBar.svelte
+++ b/src/lib/components/AudioBar.svelte
@@ -141,9 +141,11 @@ TODO:
         <PlayButton
             state={playButtonState}
             color={iconPlayColor}
-            onclick={() => {
-                checkForAudioDownload();
-                playPause();
+            onclick={async () => {
+                const canPlayPause = await checkForAudioDownload();
+                if (canPlayPause) {
+                    playPause();
+                }
             }}
         />
 
--- a/src/lib/components/AudioBar.svelte
+++ b/src/lib/components/AudioBar.svelte
@@ -141,11 +141,9 @@ TODO:
         <PlayButton
             state={playButtonState}
             color={iconPlayColor}
-            onclick={async () => {
-                const canPlayPause = await checkForAudioDownload();
-                if (canPlayPause) {
-                    playPause();
-                }
+            onclick={() => {
+                checkForAudioDownload();
+                playPause();
             }}
         />
 
--- a/src/lib/components/AudioBar.svelte
+++ b/src/lib/components/AudioBar.svelte
@@ -32,6 +32,7 @@ TODO:
     import PlayButton from './PlayButton.svelte';
     import RepeatButton from './RepeatButton.svelte';
 
+    let { checkForAudioDownload = () => {} } = $props();
     function mayResetPlayMode(hasTiming: boolean) {
         // If the current mode is repeatSelection and the reference is changed to something without timing
         // (even chapter without audio), then reset the playMode.  This matches how the Android app behaves.
@@ -137,7 +138,14 @@ TODO:
                 <AudioIcon.Replay10 color={iconColor} />
             </button>
         {/if}
-        <PlayButton state={playButtonState} color={iconPlayColor} onclick={playPause} />
+        <PlayButton
+            state={playButtonState}
+            color={iconPlayColor}
+            onclick={() => {
+                checkForAudioDownload();
+                playPause();
+            }}
+        />
 
         {#if $refs.hasAudio?.timingFile}
             <button class="audio-control-buttons" onclick={() => changeVerse(1)}>
--- a/src/lib/components/AudioBar.svelte
+++ b/src/lib/components/AudioBar.svelte
@@ -141,9 +141,11 @@ TODO:
         <PlayButton
             state={playButtonState}
             color={iconPlayColor}
-            onclick={() => {
-                checkForAudioDownload();
-                playPause();
+            onclick={async () => {
+                const audioAvailable = await checkForAudioDownload();
+                if (audioAvailable) {
+                    playPause();
+                }
             }}
         />
 
--- a/src/lib/components/AudioBar.svelte
+++ b/src/lib/components/AudioBar.svelte
@@ -32,7 +32,11 @@ TODO:
     import PlayButton from './PlayButton.svelte';
     import RepeatButton from './RepeatButton.svelte';
 
-    let { checkForAudioDownload = () => {} } = $props();
+    interface Props {
+        checkAudioAvailability?: () => Promise<boolean>; //checkAudioAvailability downloads the audio if necessary and returns true if the audio is available
+    }
+
+    let { checkAudioAvailability = () => Promise.resolve(true) }: Props = $props();
     function mayResetPlayMode(hasTiming: boolean) {
         // If the current mode is repeatSelection and the reference is changed to something without timing
         // (even chapter without audio), then reset the playMode.  This matches how the Android app behaves.
@@ -141,12 +145,12 @@ TODO:
         <PlayButton
             state={playButtonState}
             color={iconPlayColor}
-            onclick={async () => {
-                const audioAvailable = await checkForAudioDownload();
-                if (audioAvailable) {
-                    playPause();
-                }
-            }}
+            onclick={() =>
+                checkAudioAvailability().then((audioAvailable) => {
+                    if (audioAvailable) {
+                        playPause();
+                    }
+                })}
         />
 
         {#if $refs.hasAudio?.timingFile}
@@ -170,7 +174,7 @@ TODO:
             <AudioIcon.Speed color={iconColor} />
         </button>
     {/if}
-    {#if !$refs.hasAudio.timingFile}
+    {#if !$refs.hasAudio?.timingFile}
         <!-- Progress Bar -->
         <div class="audio-progress-value text-sm">
             {format($audioPlayer.progress)}
--- a/src/lib/components/AudioDownloadModal.svelte
+++ b/src/lib/components/AudioDownloadModal.svelte
@@ -21,7 +21,7 @@ Plan Stop Modal Dialog component.
         audioUrl = url;
         modal?.showModal();
     }
-    async function downloadAudio() {
+    export async function downloadAudio(url: string) {
         if (downloadAutomatically) {
             $userSettings['audio-auto-download'] = 'auto';
         }
@@ -32,8 +32,13 @@ Plan Stop Modal Dialog component.
                 book: $refs.book,
                 chapter: $refs.chapter
             },
-            audioUrl
+            url
         );
+        console.log('Download finished!'); //We'll need an actual progress bar instead of just a console.log showing when it's finished.
+        return addedAudioClip;
+    }
+    async function finishModal() {
+        const addedAudioClip = downloadAudio(audioUrl);
         if (!addedAudioClip) {
             error = 'Audio clip could not be downloaded';
             setTimeout(() => {
@@ -81,7 +86,7 @@ Plan Stop Modal Dialog component.
                 <button
                     class="dy-btn message-button pointer-events-auto"
                     id="yes"
-                    onclick={() => downloadAudio()}
+                    onclick={() => finishModal()}
                 >
                     {$t['Button_Yes']}
                 </button>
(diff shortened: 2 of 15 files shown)


src/lib/components/AudioDownloadModal.svelte
```diff
@@ -0,0 +1,157 @@
...
+    async function finishModal() {
+        const addedAudioClip = await downloadAudio(audioUrl);
+        if (!addedAudioClip && !abortController?.signal.aborted) {
+            error = 'Audio clip could not be downloaded';
+            setTimeout(() => {
+                error = '';
+            }, 2000);
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
TheNonPirate: The modal automatically closes when a button is clicked, so calling `modal?.close();` is unnecessary.

src/lib/components/AudioBar.svelte
```diff
@@ -32,6 +32,7 @@ TODO:
     import PlayButton from './PlayButton.svelte';
     import RepeatButton from './RepeatButton.svelte';
 
+    let { checkForAudioDownload = () => {} } = $props();
```
FyreByrd: Change this to 
```ts
interface Props {
    checkForAudioDownload?: () => Promise<boolean>;
}

let { checkForAudioDownload = () => Promise.resolve(true) }: Props = $props();
```

FyreByrd: You may also want to document what this means? Because you want it to return true when audio is available in general (based on the usage) not just when audio is downloaded. Maybe consider changing the name to `checkAudioAvailability`?

src/lib/components/AudioBar.svelte
```diff
@@ -137,7 +138,16 @@ TODO:
                 <AudioIcon.Replay10 color={iconColor} />
             </button>
         {/if}
-        <PlayButton state={playButtonState} color={iconPlayColor} onclick={playPause} />
+        <PlayButton
+            state={playButtonState}
+            color={iconPlayColor}
+            onclick={async () => {
```
FyreByrd: Consider changing this to
```ts
onclick={() => 
  checkForAudioDownload().then((audioAvailable) => {
      if (audioAvailable) {
          playPause();
      }
  })
}
```

src/lib/components/AudioDownloadModal.svelte
```diff
@@ -0,0 +1,157 @@
...
+<script lang="ts">
+    import { addAudioClip, updateAudioPlayer } from '$lib/data/audio';
+    import { refs, t, userSettings } from '$lib/data/stores';
+    import { CheckboxIcon, CheckboxOutlineIcon } from '$lib/icons';
+    import { tick } from 'svelte';
+    import Modal from './Modal.svelte';
+
+    let { planId = $bindable(undefined) } = $props();
+
+    const modalId = 'planStopDialog';
```
FyreByrd: I don't think this is the id you want...

src/lib/components/AudioDownloadModal.svelte
```diff
@@ -0,0 +1,157 @@
...
+        audioUrl = url;
+        modal?.showModal();
+    }
+    export async function downloadAudio(url: string) {
+        try {
+            if (downloadAutomatically) {
+                $userSettings['audio-auto-download'] = 'auto';
+            }
+            downloadProgress = 1;
+            abortController = new AbortController();
```
FyreByrd, line 30: Ooh. Not in this PR, but you may want to look into adding AbortController to some of the non-audio downloads (i.e. Verse-on-Image)

src/lib/components/AudioDownloadModal.svelte
```diff
@@ -0,0 +1,157 @@
...
+{/if}
+{#if downloadProgress > 0}
+    <div class="fixed inset-0 bg-black/50 backdrop-blur-xs flex items-center justify-center z-50">
+        <div
+            class="bg-base-100 p-6 shadow-xl w-80
+              flex flex-col justify-between h-64"
+        >
+            <div class="flex flex-col gap-1 text-left">
+                <p class="text-lg font-bold">{$t['Audio_Download_Title']}</p>
+                <!--Figure out the exact styling and such later-->
```
FyreByrd: You will need to do this before the PR can be merged...

src/lib/data/audio.ts
```diff
@@ -673,6 +676,114 @@ function getDamId(audioSource: any) {
...
+
+            if (total && onProgress) {
+                const percent = Math.round((received / total) * 100);
+                onProgress(percent);
+            }
+        }
+
+        const blob = new Blob(chunks);
+        const audioClips = await openAudioClips();
+        const date = new Date()[Symbol.toPrimitive]('number');
```
FyreByrd: Um... What does this line do???

src/lib/data/audio.ts
FyreByrd: Pretty sure this PR is the place to resolve this TODO...

src/routes/text/+page.svelte
```diff
@@ -384,6 +385,36 @@
             }
         }
     };
+    export async function checkForAudioDownload() {
```
FyreByrd: I don't think this file is the right place for this function to be defined, especially since you're exporting it?

src/lib/data/audio.ts
```diff
@@ -14,8 +14,11 @@ import {
     type Timing
 } from '$lib/data/stores';
 import { pathJoin } from '$lib/scripts/stringUtils';
+import { openDB, type DBSchema } from 'idb';
 import { logAudioDuration, logAudioPlay } from './analytics';
 
+const audioClipUrls = new Map<string, string>();
```
FyreByrd: Please use MRUCache instead of Map for this use case.

src/lib/data/audio.ts
FyreByrd, line 1: Please separate the IndexedDB stuff into a separate file (I eventually have plans to move all the IndexedDB stuff to a subfolder)

FyreByrd (changes requested):
I found a styling bug in Modal.svelte when switching between themes that you could fix here.

Change the source of Modal.svelte to
```svelte
<!--
@component
A simple modal component from DaisyUI. Closes when clicked outside of.

See https://daisyui.com/components/modal/#modal-that-closes-when-clicked-outside
@prop { String }   id      - ID for the modal.
@prop { Function } content - Snippet containing the content of the modal.
@prop { String }   addCSS  - CSS to inject into the model contents div/form.
@prop { Function } onclose - Function to run when Modal closes.
-->
<script lang="ts">
    import { direction, themeColors } from '$lib/data/stores';
    import type { Snippet } from 'svelte';

    interface Props {
        id: string;
        children?: Snippet;
        styling?: string;
        onclose?: () => void;
        dialog?: HTMLDialogElement;
    }

    let { id, children, styling, onclose, dialog = $bindable() }: Props = $props();
    /**
     * This exported function allows buttons/labels
     * in other divs to trigger the modal popup
     */
    export function showModal() {
        dialog?.showModal();
    }
</script>

<dialog
    bind:this={dialog}
    {id}
    {onclose}
    class="dy-modal cursor-pointer"
    style:direction={$direction}
>
    <form
        method="dialog"
        style={styling ?? `background-color:${$themeColors['DialogBackgroundColor']};`}
        class="dy-modal-box overflow-y-visible relative"
    >
        {@render children?.()}
        <!--This is the snippet for the popup's actual contents-->
    </form>
    <form method="dialog" class="dy-modal-backdrop">
        <!--This allows the modal to be closed when the user taps outside of the contents div/form-->
        <button>close</button>
    </form>
</dialog>
```

src/routes/text/+page.svelte
TheNonPirate: Where would be a better place to define it? I realized it doesn't actually need to be exported; I think it was exported because a previous iteration of it needed to be exported.

TheNonPirate:
Do we have a way to test FCBH Bible Brain audio?

src/lib/data/audio.ts
TheNonPirate: Won't that lead to the object URLs being retained when they're evicted from the MRUCache, meaning that it won't achieve its goal of avoiding too unnecessary object URLs being created? I'd think it would be much more worth it to have a complete map of strings rather than to potentially leak some blob object URLs that could be much larger in size.

FyreByrd: Retained where?

FyreByrd:
> Do we have a way to test FCBH Bible Brain audio?

Add an audio source in the test project in SAB. Unfortunately the current FCBH audio source is semi-broken, not sure how that happened...

src/routes/text/+page.svelte
FyreByrd: You could probably put it in audio.ts?

TheNonPirate:
There's still an issue that I haven't figure out yet, which may be with Bible Brain audio files specifically, or may just be something that I noticed while testing them:
When I click between different chapters of Matthew after replacing Matthew's audio files with Bible Brain ones, then choose random chapters to click play on, sometimes after downloading a chapter and playing, it will play the audio file that goes with an adjacent chapter rather than that chapter. I'm having trouble reproducing this issue because I can't figure out exactly what triggers it.
EDIT: It seems that it doesn't download the wrong file; it plays the wrong file, because when I reload the page, it plays the correct file. I'm wondering if it's related to the MRUCache.
EDIT 2: It seems that when the issue occurs, then if I keep switching to the next chapter and downloading its audio and playing it, it also plays the audio for the wrong chapter (If I am on Chapter 11 and it plays Chapter 12, then if I go to Chapter 12, it plays Chapter 13).

TheNonPirate:
It seems this bug also exists on main, and I think it's caused by an audio load that doesn't finish until after the chapter is switched, causing it to start playing the wrong chapter's audio, which then cannot be cancelled once the correct audio loads.

FyreByrd (changes requested):
Works very well. The only thing missing is the no connection popup.

Image

FyreByrd:
> Works very well. The only thing missing is the no connection popup.

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine

TheNonPirate:
The cache seems to cause some problems, such as not prompting the user to download audio because the cache is updated even if the audio is not downloaded. With the new cache check, I am finding myself unable to get the download modal to pop up.

FyreByrd:
> The cache seems to cause some problems, such as not prompting the user to download audio because the cache is updated even if the audio is not downloaded. With the new cache check, I am finding myself unable to get the download modal to pop up.

Ok. It can be removed then. Sorry about that.

TheNonPirate:
The docs for navigator.onLine say, "Therefore, this property is inherently unreliable, and you should not disable features based on the online status, only provide hints when the user may seem offline." Right now, I'm using navigator.onLine, but would it be better not to use it, or to somehow include a fallback in case it thinks the user is offline when they aren't?

FyreByrd:
> The docs for navigator.onLine say, "Therefore, this property is inherently unreliable, and you should not disable features based on the online status, only provide hints when the user may seem offline." Right now, I'm using navigator.onLine, but would it be better not to use it, or to somehow include a fallback in case it thinks the user is offline when they aren't?

I'll whip something up real quick

FyreByrd:
I've added a *slightly* more sophisticated connection checker. It may be useful to check the connection before calling getBibleBrainUrl on line 897 of audio.ts...

FyreByrd (approved):

PR #1079, opened 2026-07-22
Fix collections without scripture books

### LINKED ISSUE #581: Book Collection without traditional books breaks conversion process
If one of the book collections in an SAB project has no traditional books (e.g. it only has quizzes), then the conversion process will fail during the convert books step.

The failure happens because the Proskomma instance for the book collection does not contain any documents, and thus does not return usable information when queried during the catalog creation process.

PR #580 attempts to address this issue, but potentially causes much bigger problems elsewhere.

Fixing this issue may require a fix in SAB to disallow collections without traditional books, unless that is a valid use case.

TheNonPirate:
Fixes #581. This PR generates dummy Proskomma data for collections without Scripture books so that the convert process does not fail, as well as tweaking some runtime behavior to account for collections without Scripture books. It also makes the text route redirect to the appropriate quiz route if it detects that the current book is a quiz.

## Summary by CodeRabbit

* **New Features**
  * Quiz-based books now automatically open the corresponding quiz experience from the text page.
* **Bug Fixes**
  * Improved active-book switching when navigating between books and collections.
  * Improved catalog generation for empty results to ensure consistent, valid output.
  * Prevent writing frozen archives with missing content by defaulting to an empty value.
* **Improvements**
  * More consistent book format selection and illustration loading based on the active collection/book.
  * Enhanced quiz page header/navigation, including a conditional collection badge and back button.

### DIFF
--- a/src/lib/data/navigation.ts
+++ b/src/lib/data/navigation.ts
@@ -174,7 +174,7 @@ export class NavigationContext {
 
     private async updateLocation(docSet: string, book: string, chapter: string, verse: string) {
         let newBook = await this.updateCatalog(docSet);
-        if (book !== this.book && this.allBookIds.includes(book)) {
+        if (book !== this.book && (this.allBookIds.includes(book) || !this.book)) {
             this.book = book;
             newBook = true;
         }
--- a/src/lib/data/navigation.ts
+++ b/src/lib/data/navigation.ts
@@ -162,9 +162,15 @@ export class NavigationContext {
             this.collection = docSet.split('_')[1];
             this.catalog = await this.fetchCatalog(this.docSet);
             this.books = this.catalog.documents.map((b) => b.bookCode);
+            const collection = this.config.bookCollections?.find(
+                (c) => docSet === c.id || docSet === this.docSetFromCollection(c)
+            );
+            const quizzes =
+                collection?.books.filter((x) => x.type === 'quiz').map((x) => x.id) || [];
             this.allBookIds = [
                 ...this.books,
-                ...(this.catalog.htmlBooks ? this.catalog.htmlBooks.map((b) => b.id) : [])
+                ...(this.catalog.htmlBooks ? this.catalog.htmlBooks.map((b) => b.id) : []),
+                ...quizzes
             ];
             return true;
         } else {
@@ -174,8 +180,7 @@ export class NavigationContext {
 
     private async updateLocation(docSet: string, book: string, chapter: string, verse: string) {
         let newBook = await this.updateCatalog(docSet);
-
-        if (book !== this.book) {
+        if (book !== this.book && this.allBookIds.includes(book)) {
             this.book = book;
             newBook = true;
         }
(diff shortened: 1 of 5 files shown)


FyreByrd (changes requested):
I created a collection with only quizzes. The back button does not work. There is also no collection selector. I am completely stuck in the quiz collection now.

I would also take a closer look at the CodeRabbit feedback, as it looks reasonable.

TheNonPirate:
> I created a collection with only quizzes. The back button does not work. There is also no collection selector. I am completely stuck in the quiz collection now.

Was this done in an app without a contents page? When I try it in a modified form of the SAB Test PWA, the back button returns me to the contents page.

EDIT: I was able to reproduce it in an app without a contents page.

FyreByrd:
> Was this done in an app without a contents page? When I try it in a modified form of the SAB Test PWA, the back button returns me to the contents page.

Yeah. No contents page. I did the test with the WEB Gospels project

src/lib/data/navigation.ts
```diff
@@ -174,7 +174,8 @@ export class NavigationContext {
 
     private async updateLocation(docSet: string, book: string, chapter: string, verse: string) {
         let newBook = await this.updateCatalog(docSet);
-        if (book !== this.book && this.allBookIds.includes(book)) {
+
```
FyreByrd: If you can, I think a better fix would be to make sure that quizzes (and also songbooks and storybooks?) are included in allBookIds.

TheNonPirate: Done. It seems that all other types of books are included in the catalog and thus in allBookIds. Quizzes are the only different ones. Do you know why quizzes are handled differently?

src/routes/quiz/[collection]/[id]/+page.svelte
```diff
@@ -291,6 +304,20 @@
             {/snippet}
         </Navbar>
     </div>
+    {#if showCollectionViewer && enoughCollections}
```
Comment (FyreByrd) [inline src/routes/quiz/[collection]/[id]/+page.svelte]: Is this present in the native app???

Comment (TheNonPirate) [reply src/routes/quiz/[collection]/[id]/+page.svelte]: Yes. It just wasn't in the PWA for some reason.

Comment (FyreByrd) [reply src/routes/quiz/[collection]/[id]/+page.svelte]: Ok... Because I'm not seeing it in the native app on my end...

Comment (TheNonPirate) [reply src/routes/quiz/[collection]/[id]/+page.svelte]: This is what it looks like in the native app on my end.
image

src/lib/data/navigation.ts
FyreByrd: Probably because they were the first to be an entirely different route?

src/routes/quiz/[collection]/[id]/+page.svelte
Comment (FyreByrd) [reply src/routes/quiz/[collection]/[id]/+page.svelte]: What version of SAB are you running? The most recent version is supposed to have a status bar showing where you are in the quiz (not that that needs to be implemented in the PWA yet...)

Comment (FyreByrd) [reply src/routes/quiz/[collection]/[id]/+page.svelte]: The status bar is definitely hiding the collection selector, assuming it's even still there...

Screenshot 2026-07-24 at 9 22 54 AM

Comment (FyreByrd) [reply src/routes/quiz/[collection]/[id]/+page.svelte]: The collection selector does not appear to be there, but there is a hamburger menu instead of the back button, and this includes a Language & Layout option, that would allow selecting a collection.

FyreByrd (approved):

PR #1084, opened 2026-07-27
Allow navigating to intros using swipes or arrows

### LINKED ISSUE #1083: Cannot access intro by navigating left
In the native app, if the book the user is in has an intro, swiping to navigate left takes the user to the intro, and swiping to navigate right from the intro takes the user to the first chapter. In the PWA, the right arrow appears on the intro to navigate to the first chapter, but no left arrow appears on the first chapter to navigate to the intro.

TheNonPirate:
Fixes #1083. In the native app, you could access a book's introduction by navigating left when on chapter 1, but in the PWA, this is not possible, and if 'Allow swiping between books' is set, the introductions will be skipped when swiping between books. To fix this, I made $refs.next and $refs.prev get set to book introductions if an intro should be the next or previous chapter.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved next/previous navigation across book boundaries when a book has an introduction.
  * Navigation now opens the introduction when it’s available, including correct handling at the first and last chapter edge cases.
  * Prevented invalid “previous” navigation when there is no earlier content to go back to.
  * Updated navigation tests to reflect introduction chapter behavior.

### DIFF
--- a/src/lib/data/navigation.test.ts
+++ b/src/lib/data/navigation.test.ts
@@ -507,7 +507,7 @@ describe('goTo', () => {
             const navContext = new TestNavigationContext(getTestCatalog, config);
             await navContext.gotoInitial();
             await navContext.goto('eng_C01', 'MAT', '3', '1');
-            const next = { book: 'MRK', chapter: '1' };
+            const next = { book: 'MRK', chapter: 'i' };
             expect(navContext.next).toStrictEqual(next);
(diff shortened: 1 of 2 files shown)


TheNonPirate:
It looks like I'll need to update some tests.

FyreByrd (approved):

PR #1085, opened 2026-07-27
Update bottom navbar icon color and font

### LINKED ISSUE #1064: Bottom navigation bar items don't use correct font or icon highlight color
PWA/Android:

Image

Image

TheNonPirate:
Fixes #1064. This PR updates the bottom navigation bar icons to be colored based on the color the app defines. It also updates the bottom navbar item text to use the system font like the default for the PWA. Right now, it doesn't directly use the ui.bottom-navigation.item.text class; it just gets the color from the defined classes, so this'll need to be updated if we want to support arbitrary styles or changing the font. Is that within the scope of this issue?

## Summary by CodeRabbit

- **Style**
  - Updated the bottom navigation bar icons to use CSS mask-based rendering for more consistent visuals across themes.
  - Improved the selected-state look by adjusting icon/background emphasis and refining how the label styling switches between active and inactive states.

### DIFF
--- a/src/lib/components/BottomNavigationBar.svelte
+++ b/src/lib/components/BottomNavigationBar.svelte
@@ -135,19 +135,17 @@
                             class="dy-btn dy-btn-ghost flex-col gap-1 my-2"
                             onclick={() => handleClick(item.type, item.link?.['default'])}
                         >
-                            <picture>
-                                <!-- Image Icon -->
-                                <div
-                                    style={`width:24px; height:24px; background: ${
-                                        selectedLink(item.type, item.link?.['default'])
-                                            ? barTextSelectedColor
-                                            : barTextColor
-                                    }; mask: url(${menuIcons[`./${item.images?.[0]?.file}`]}); -webkit-mask: url(${menuIcons[`./${item.images?.[0]?.file}`]});`}
-                                    class={selectedLink(item.type, item.link?.['default'])
-                                        ? 'opacity-100'
-                                        : 'opacity-50'}
-                                ></div>
-                            </picture>
+                            <!-- Image Icon -->
+                            <div
+                                style={`width:24px; height:24px; background: ${
+                                    selectedLink(item.type, item.link?.['default'])
+                                        ? barTextSelectedColor
+                                        : barTextColor
+                                }; mask: url(${menuIcons[`./${item.images?.[0]?.file}`]}); -webkit-mask: url(${menuIcons[`./${item.images?.[0]?.file}`]});`}
+                                class={selectedLink(item.type, item.link?.['default'])
+                                    ? 'opacity-100'
+                                    : 'opacity-50'}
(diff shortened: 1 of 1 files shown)


FyreByrd:
> Right now, it doesn't directly use the ui.bottom-navigation.item.text class; it just gets the color from the defined classes, so this'll need to be updated if we want to support arbitrary styles or changing the font. Is that within the scope of this issue?

I think this would be close enough to the PR's scope. It also fits with the overall goal of gradually cleaning up how we handle styles. If you can do it in around 15 or so minutes, I think that would be great.

FyreByrd (approved):
This is fine as is. If you want, you can fix the class usage as discussed above.

TheNonPirate:
As far as I can tell, the ui.bottom-navigation.item.text class isn't written as a class into the sab-app.css file and thus probably can't be used as a class. Am I wrong, and if not, is there an easy way to still apply its styles to the text elements?

FyreByrd:
> As far as I can tell, the ui.bottom-navigation.item.text class isn't written as a class into the sab-app.css file and thus probably can't be used as a class. Am I wrong, and if not, is there an easy way to still apply its styles to the text elements?

convertStyle with the `$s` store... (as in TabsMenu.svelte and many other components...)

src/lib/components/BottomNavigationBar.svelte
```diff
@@ -22,8 +22,8 @@
     const barBackgroundColor = $derived(
         ($s?.['ui.bottom-navigation.bar'] ?? $s?.['ui.bottom-navigation.'])?.['background-color']
     );
-    const barTextColor = $derived($s?.['ui.bottom-navigation.item.text']['color']);
-    const barTextSelectedColor = $derived($s?.['ui.bottom-navigation.item.text.selected']['color']);
+    const barTextColor = $derived($s?.['ui.bottom-navigation.item.icon']['color']);
```
FyreByrd: Bikeshedding here, but please rename this to barIconColor, and similar for below.

src/lib/components/BottomNavigationBar.svelte
```diff
@@ -142,21 +142,21 @@
...
                                     'font-family'
-                                ] ?? 'system-ui'}; color: {selectedLink(
-                                    item.type,
-                                    item.link?.['default']
-                                )
-                                    ? barTextSelectedColor
-                                    : barTextColor}"
+                                ]
+                                    ? ''
+                                    : 'system-ui'}; {convertStyle(
```
FyreByrd: You don't need to do this extra check for if the font-family is there or not. You should be able to just set it to 'system-ui' to start, and then let CSS handle the logic if font-family happens to be set in the config styles.

FyreByrd (changes requested):

FyreByrd (approved):

PR #1089, opened 2026-07-29
Implement Highlight Words tab in Verse on Image

### LINKED ISSUE #1011: Verse on Image: Edit color and size of single words by selection
There is now a new “Highlight Words” tab in the Verse on Image editor, next to the Text Color tab.

Image

TheNonPirate:
Resolves #1011. This PR implements the Highlight Words tab in Verse on Image, which allows individual words to be colored and their case to be set. It also fixes an issue with Verse on Image where the carousel of tabs can be scrolled using the arrow keys, but it does not fully switch tabs.

## Summary by CodeRabbit

* **New Features**
  * Added word-level selection and editing for verse overlays, including per-word colors and locale-aware uppercase/lowercase controls.
  * Introduced a dedicated Word Editor Pane and new word-selection and navigation icons.

* **Refinements**
  * Improved editor tab alignment and pane scrolling for smoother navigation.
  * Updated download controls with a cleaner, borderless appearance and improved contrast.

* **Bug Fixes**
  * Corrected editor pane labels and text alignment wording.


FyreByrd (changes requested):
The functionality is all there and works just like the native app as far as I can tell. Well done.

However, there are a few UI issues...

Image

Image

Also, there are tabs that should not need to be vertically scrollable (e.g. the alignment tab) but are vertically scrollable. I think this is because the container for the tabs content has its height set by the tallest element. It looks like this may be a limitation of the DaisyUI classes being used. If you can fix it without much effort, great. If not, it's probably not a big deal.

TheNonPirate:
How did you get those issues, particularly the vertical scrolling? I'm having trouble reproducing them.

FyreByrd:
> How did you get those issues, particularly the vertical scrolling? I'm having trouble reproducing them.

How did I get the issues? By running the code... I didn't do anything special to get those issues...

TheNonPirate:
What device/browser were you running it on? I can only get the first issue by making the width of the page really small, and I can't get the second issue.

FyreByrd:
> What device/browser were you running it on? I can only get the first issue by making the width of the page really small, and I can't get the second issue.

Brave Browser. I'll check the screen dimensions rq

FyreByrd:
> > What device/browser were you running it on? I can only get the first issue by making the width of the page really small, and I can't get the second issue.
> 
> Brave Browser. I'll check the screen dimensions rq

Ok. So the max height of my browser window is 638 pixels. The width of the container depends on the width of the image. I am using SAB Test PWA. It shouldn't matter which verse. I am using the first image, which has a width of 375px. Also, another weird bug, the image seems to disappear when the width transitions across 450px and does not reappear until a new image is selected...

TheNonPirate:
Do you have a screenshot of the vertical scrolling issue?
Should I try to fix the image disappearing when the width gets too small, or is that not an issue because changing the screen width is mostly a development thing?

FyreByrd:
> Do you have a screenshot of the vertical scrolling issue? Should I try to fix the image disappearing when the width gets too small, or is that not an issue because changing the screen width is mostly a development thing?

Screenshot 2026-07-30 at 11 29 24 AM
Screenshot 2026-07-30 at 11 29 35 AM

I think the resize issue is mostly a development thing, so I honestly wouldn't worry about it.

src/lib/components/VerseOnImage.svelte
```diff
@@ -981,12 +981,22 @@ The verse on image component.
         </div>
 
         <!-- Font Selector Pane -->
-        <div class="dy-carousel-item items-center editorPane">
+        <div
+            class="dy-carousel-item items-center editorPane"
+            style="
+                overflow-y: auto;
```
FyreByrd: Uh... this could be a tailwind class instead.... try `overflow-y-auto`...

FyreByrd: In fact, if this is a style that is meant to apply to every editor pane, maybe you could put it in the `.editorPane` class...

FyreByrd (changes requested):
I think it would be better UX to have the case change buttons located above the color picker, but that may be just me...

src/lib/components/DownloadSelector.svelte
```diff
@@ -26,18 +26,20 @@ A component for verse-on-image providing a dropdown where you can choose to down
...
+            style="background-color:{$themeColors['PopupBackgroundColor']};"
             onclick={() => downloadImage()}
         >
             <ImageIcon.Image />
             {$t['Text_On_Image_Save_Image']}
         </button>
         <button
-            class="dy-btn dy-btn-sm flex items-center justify-center gap-2"
+            class="dy-btn dy-btn-sm flex items-center justify-center gap-2 border-none text-gray-700"
+            style="background-color:{$themeColors['PopupBackgroundColor']};"
```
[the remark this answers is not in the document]
TheNonPirate: The native app seems to just have a hardcoded text color, so it does not have good contrast in the dark theme. I don't know whether this is something we want to change for the PWA.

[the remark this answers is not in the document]
FyreByrd: @TheNonPirate If the native app has poor contrast, that is an issue that probably needs to be fixed in the native app. This is one of the cases where you shouldn't blindly follow the native app...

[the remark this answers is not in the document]
TheNonPirate: What color should I use for it? Should I invert the background color to get the text color?

src/lib/components/VerseOnImage.svelte
```diff
@@ -817,7 +838,21 @@ The verse on image component.
...
-                {txtFormatted}
+                {#each words as word, i}
+                    <!-- svelte-ignore a11y_click_events_have_key_events -->
+                    <span
+                        onclick={() => {
+                            if (active_editor_index === wordEditorIndex) {
+                                word.color = `color:${wordColor};`;
+                                wordSelected = word;
+                            }
+                        }}
```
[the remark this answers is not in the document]
TheNonPirate: Words are supposed to have their color updated when selected.

src/lib/components/DownloadSelector.svelte
```diff
@@ -22,18 +22,22 @@ A component for verse-on-image providing a dropdown where you can choose to down
...
-    styling="background-color:{$themeColors['PopupBackgroundColor']}; width:auto; {positioningCSS}"
+    styling="padding:0; background-color:{$themeColors[
+        'PopupBackgroundColor'
+    ]}; width:auto; {positioningCSS}"
 >
     <div class="grid gap-2 m-2">
         <button
-            class="dy-btn dy-btn-sm flex items-center justify-center gap-2"
+            class="dy-btn dy-btn-sm flex items-center justify-center gap-2 border-none text-gray-700"
+            style="background-color:{$themeColors['PopupBackgroundColor']};"
```
FyreByrd: in lieu of setting this here, you could use dy-btn-ghost?

FyreByrd: Also for the text color, you could probably get by with `TextColor`

TheNonPirate: What's the point of using dy-btn-ghost, which has some other irrelevant effects as well, rather than just using border-none?

FyreByrd: One of those "irrelevant" effects is making the background transparent... meaning you don't have to set the background on each individual button if you already have it set on the container for the buttons. It also sets the text color to inherit, meaning you could set that on the container as well. Are any of the "irrelevant" effects problematic??

FyreByrd: Also, you don't need `items-center`, `justify-center`, or `gap-2`, since `dy-btn` appears to accomplish the same thing...

FyreByrd (approved):

PR #1093, opened 2026-08-03
fix: fall back to HTTPS when HTTP audio downloads fail

chrisvire:
Public deployments configure http:// audio URLs that get redirected to https://, but the redirect response lacks CORS headers, causing fetch() to fail outright. MicroPi/offline deployments serve audio over plain HTTP and must keep working unchanged, so URLs can't be unconditionally rewritten to https://.

addAudioClip() now goes through fetchWithProtocolFallback(), which tries the configured HTTP URL first and only retries over HTTPS if the request fails before producing a response. A successful HTTPS origin is remembered in memory for the rest of the session so later downloads skip straight to HTTPS, without persisting a determination that could go stale across environments.

## Summary by CodeRabbit

- **Bug Fixes**
  - Audio clip requests now automatically retry over HTTPS when an HTTP request fails.
  - Successful HTTPS fallback preferences are remembered for each origin.
  - Abort errors and other request failures continue to be handled correctly.

- **Tests**
  - Added coverage for fallback behavior, HTTPS requests, error propagation, and independent origin preferences.

### DIFF
--- a/src/lib/data/audioclipsDB.test.ts
+++ b/src/lib/data/audioclipsDB.test.ts
@@ -38,10 +38,7 @@ describe('fetchWithProtocolFallback', () => {
         // Subsequent request to the same origin should go straight to HTTPS.
         const secondHttpsResponse = new Response('ok');
         (fetch as any).mockResolvedValueOnce(secondHttpsResponse);
-        const secondResponse = await fetchWithProtocolFallback(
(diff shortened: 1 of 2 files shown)


FyreByrd (approved):
I don't really have a good way to test this, but I suppose the included test is enough for verification purposes

PR #1094, opened 2026-08-03
Fix audio download bug

TheNonPirate:
If the audio access method is set to download without streaming being an option, it will now download the audio instead of relying on the user setting of whether to download or stream.

## Summary by CodeRabbit

* **New Features**
  * Improved audio access handling when downloads are required.
  * Supports offline access and provides download options when streaming is unavailable.
  * Existing downloaded audio remains immediately accessible.

* **Bug Fixes**
  * Prevented unnecessary download prompts for streaming-capable audio sources.


TheNonPirate:
I'm having some issues resolving the merge conflicts, so it's going to be easier to just remake this branch (Because the changes are so small) instead of figuring out the merge conflicts.

PR #1095, opened 2026-08-03
Fix various download/sharing related issues

TheNonPirate:
Takes work from #1094, but will include other fixes as well. Current fixes:

- If the audio access method is set to download without streaming being an option, it will now download the audio instead of relying on the user setting of whether to download or stream.
- If the user attempts to share remote audio while there is no network connection, display the Audio_Download_Connect error.
- Clicking the play button in the text selection toolbar triggers audio download functionality just like clicking the normal play button or the audio on/off button.

## Summary by CodeRabbit

* **New Features**
  * Improved audio access for downloadable content.
  * Added support for offline audio access with appropriate alerts.
  * Added download prompts and automatic downloads when streaming is unavailable.
  * Downloaded audio is now stored and retrieved consistently for playback.

* **Bug Fixes**
  * Existing downloaded audio bypasses unnecessary download prompts.
  * Improved handling of unavailable audio and download links.
  * Download completion and playback updates now behave more reliably.

### DIFF
--- a/src/lib/components/AudioAlertModal.svelte
+++ b/src/lib/components/AudioAlertModal.svelte
@@ -2,6 +2,12 @@
 @component
 No Connection Modal Dialog component.
 -->
+<script lang="ts" module>
+    export interface AudioAlertModalProps {
+        messageKey: string;
+        details?: string;
+    }
+</script>
 
 <script lang="ts">
     import { t } from '$lib/data/stores';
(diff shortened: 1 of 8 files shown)


FyreByrd (approved):
Looks good to me. Just a few quick tweaks I've added, if you don't mind looking them over.

TheNonPirate:
I don't see any problems with it.

PR #1097, opened 2026-08-03
fix: play audio automatically once a required download finishes

chrisvire:
Pressing play on undownloaded audio would kick off the download but never resume playback afterward, forcing the user to tap play again. updateAudioPlayer() now accepts an autoplay option that triggers play() once the newly downloaded audio finishes loading.


FyreByrd (approved):
LGTM! works fine on my end

FyreByrd:
I'm going to go ahead and merge, since this shouldn't mess with any ongoing changes.

Issue #1092, opened 2026-08-03
Error Message: missing color assignment


chrisvire:
I think this is trying to display an error message:

<img width="1392" height="1522" alt="Image" src="https://github.com/user-attachments/assets/e3c52f3d-b8b7-428c-b4ed-a57f294fff38" />

I ran a release build and this was the message:

<img width="1392" height="1522" alt="Image" src="https://github.com/user-attachments/assets/9cc50024-c603-4cae-95f1-6858a5856d95" />

Also, the error message should not say "audio clip" ???

PR #1101, opened 2026-08-04
Handle missing colors in a way that will allow the app to build with warnings

chrisvire:
In #1092, we chose to fail the build when a color value is missing. In SAB in the desktop and the native app, it will use the value of the Normal theme with a color value is missing. Implement that strategy and give a warning when this happens so that the user know how to fix the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of deprecated theme colors.
  * Missing colors now fall back to the `Normal` theme when available.
  * Added clearer warnings when fallback values are used.
  * Errors are reported only when no suitable theme value exists.


FyreByrd (approved):
LGTM

Issue #1086, opened 2026-07-29
Verse on Image: Support new text area configuration


davidmoore1:
Verse on Image has allowed for a new default text area to be defined.  The area to be used is set on a by image basis.  Each may or may not be configured to have the text area defined.  A sample configuration of the background images with the new values set for the first is:
```
    <images type="background">
        <image width="1080" height="1080" text-area="0.098 0.098 0.782 0.480">rocio-guillen-bmVs4mDwI3k-unsplash-1080.jpg</image>
        <image width="1080" height="1080">chris-gallimore-f9fJ6nxndoo-unsplash-1080.jpg</image>
        <image width="1080" height="1080">decorative-border-dv167.jpg</image>
        <image width="1080" height="1080">sam-ueGaQiHV86o-unsplash-1080.jpg</image>
    </images>
```
The text area fields are a decimal percentage of the screen to be used as a margin in the order left, top, width, height.  So in the above configuration, I originally used the SAB editor set it to 10% left 12% right 10% top and 12% bottom and the above is what I got: roughly 10% left, 10% top, 78% width, 48% height.  This basically means that when this image is selected, the text should appear centered within the top half of the image with small margins right and left.  Below is a picture of what this looks like when you select it initially.

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/8b0d244c-bbd5-4322-8fa4-68f23b2538e2" />

A few notes:
1) Not all images have the text area defined.  When it is not, use the whole area as it is currently implemented.
2) The text font size should be adjusted when the image with the text area is displayed so that the text fits within the defined text area.  
3) This is initial placement only for when the image is selected.  The user may drag the text to any part of the screen they wish, reset the font size so it takes more of the screen.  The text area is not a restriction on where the user may place the text.  It is simply the initial placement.  
4) When you switch from one image to another, the text places itself in the text area defined for that image and to the center if the image does not have a text placement defined.

FyreByrd:
Closed by #1103

PR #1103, opened 2026-08-04
Implement text area configuration for verse on image

TheNonPirate:
Implements #1086. This PR uses text area data from verse on image backgrounds to set the textbox positioning and size when switching to a background.

## Summary by CodeRabbit

* **New Features**
  * Added support for defining custom text areas on background images.
  * Text overlays now position and center within configured areas when available.

* **Bug Fixes**
  * Improved text box sizing during touch and pointer resizing.
  * Prevented text from displaying incorrectly while its dimensions are being calculated.
  * Improved synchronization between text width controls and the displayed text box.

### DIFF
--- a/config/index.d.ts
+++ b/config/index.d.ts
@@ -300,6 +300,7 @@ export type ScriptureConfig = AppConfig & {
         width: string;
         height: string;
         filename: string;
+        textArea?: { left: number; top: number; width: number; height: number };
     }[];
     watermarkImages?: {
         width: string;
--- a/convert/convertConfig.ts
+++ b/convert/convertConfig.ts
@@ -1670,7 +1670,21 @@ export function parseBackgroundImages(document: Document, verbose: number) {
             const width = backgroundImage.getAttribute('width')!;
             const height = backgroundImage.getAttribute('height')!;
             const filename = backgroundImage.innerHTML;
-            backgroundImages.push({ width, height, filename });
+            const textAreaString = backgroundImage.getAttribute('text-area');
+            let textArea: { left: number; top: number; width: number; height: number } | undefined =
+                undefined;
+            if (textAreaString) {
+                const textAreaArray = textAreaString.split(' ');
+                if (textAreaArray.length === 4) {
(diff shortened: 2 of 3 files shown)


FyreByrd (changes requested):
I can drag the text outside of not only the bounding box, but the image itself...

Image

TheNonPirate:
I think that's still within the image. That image has a lighter-colored border at the edge of it, and I'm pretty sure the words are just on that rather than outside of the image. Being able to move the text outside of the text area configured in SAB is intentional; the text area configuration just determines the default positioning of the text.

FyreByrd:
> I think that's still within the image. That image has a lighter-colored border at the edge of it, and I'm pretty sure the words are just on that rather than outside of the image. Being able to move the text outside of the text area configured in SAB is intentional; the text area configuration just determines the default positioning of the text.

Native:

https://github.com/user-attachments/assets/d310924e-3e71-48e0-a529-86614125de83

PWA:

https://github.com/user-attachments/assets/5c10aab3-11ca-4ee3-a7f7-76ce22c7d676

Should this be a separate issue then?

TheNonPirate:
Maybe it should be a separate issue. Do you have any idea what the exact margins are in the native app?

FyreByrd:
> Maybe it should be a separate issue. Do you have any idea what the exact margins are in the native app?

Is this what you're asking?
Screenshot 2026-08-07 at 11 46 22 AM

TheNonPirate:
The text area does not constrain where the user can move the text. I have set the text area to this area and yet been able to move it just as far throughout the image as when I didn't change the text area.

image

FyreByrd (approved):

TheNonPirate:
Did you decide that the margin issue just isn't big enough to worry about, or that it should be a separate issue?

FyreByrd:
> Did you decide that the margin issue just isn't big enough to worry about, or that it should be a separate issue?

Honestly it's probably fine. It could be opened as a separate issue, but it's not really high priority

PR #1108, opened 2026-08-06
Fix settings styling

### LINKED ISSUE #1091: Settings: missing color assignment
We need to figure out how the native app is styling some of the items in the settings:

Image

TheNonPirate:
Resolves #1091. This PR replaces the remaining DaisyUI color styling in Settings.ts with styling that better matches the native app. `<select>` components are difficult to style and some of the styling is simultaneously div-only and not in $s, so I had to just use the default SAB colors (As in, the ones that are in the correct style in the SAB Test PWA, which I assume is at least used in the vast majority of SAB projects because of the specificity of the color names) for part of it rather than using the SAB classes.

## Summary by CodeRabbit

* **Style**
  * Improved settings controls with theme-aware checkbox colors and selector styling.
  * Updated list selectors with clearer focus states and themed menus for improved readability.
  * Enhanced settings control presentation across light and dark themes for a more consistent visual experience.

### DIFF
--- a/src/lib/components/Settings.svelte
+++ b/src/lib/components/Settings.svelte
@@ -2,7 +2,6 @@
     import config from '$assets/config';
     import {
         language,
-        s,
         SettingsCategory,
         t,
         theme,
@@ -27,7 +26,6 @@
         config.interfaceLanguages?.writingSystems[$language]?.fontRelativeSize
     );
     const fontSize = $derived(fontRelativeSize ? fontRelativeSize : '100');
-    console.log($s);
 </script>
 
 <!-- loops through the different settings types -->
--- a/src/lib/components/Settings.svelte
+++ b/src/lib/components/Settings.svelte
@@ -2,6 +2,7 @@
     import config from '$assets/config';
     import {
         language,
+        s,
         SettingsCategory,
         t,
         theme,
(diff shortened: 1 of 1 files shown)


src/lib/components/Settings.svelte
```diff
@@ -63,15 +72,20 @@
...
-                </select>
+                <div class="settings-summary">
+                    <select
+                        class="dy-select dy-select-ghost dy-select-sm appearance-none px-0 w-full outline-none settings-select"
+                        style:font-size="{fontSize}%"
+                        bind:value={$userSettings[setting.key]}
+                    >
+                        {#each setting.entries ?? [] as entry, i}
+                            <option
+                                style="background: var(--DialogBackgroundColor); color: var(--SettingsTitleColor);"
```
FyreByrd: It might be better to do this as
```
style:background-color={$themeColors['DialogBackgroundColor]}
style:color={$themeColors['SettingsTitleColor']}
```
I know we haven't been consistent about this in the past.

Even if you don't do it this way, `background-color` is the CSS property you should be using here, not `background`.

FyreByrd (changes requested):
Looks good overall, just a small fix needed

FyreByrd (approved):

PR #1117, opened 2026-08-10
Implement Download Manager

### LINKED ISSUE #1036: Download audio for offline use

### LINKED ISSUE #1112: Implement download manager
The download manager is an item off of the slide drawer that manages audio files for books that can be downloaded.  Typically, audio files have been downloaded as the chapters they were associated with were being viewed.  Once downloaded, the files stayed in the cache until the app was deleted.   With the download manager, the user can select to have part or all of the downloadable audio files associated with the books to be loaded at once, which can be useful if there is limited access to the internet.  It also allows the user to delete any files that were not packaged with the app to be removed.

The option is controlled by two parameters.  The first is in the "features" section of the appDef file called "audio-download-manager".  It defaults to true, so if it is not present in the appDef file, the feature is enabled.  The other is in the "traits" section of the appDef file and is named "has-downloadable-audio".  It has a true or false value.  If the feature is enabled and the app has downloadable audio, the "Downloads" option is displayed in the slide menu as shown below.
Image

The initial display depends on whether there is more than one book collection defined with audio files that are either included as assets or downloadable.   From here on down "downloadable audio" will mean all of the asset chapter files plus all the downloadable chapter files.  Stream only audios are not included in any way in the download manager.   If there are multiple book collections with downloadable audio or asset files, then a book collection selection screen is displayed.  If there is only one book collection, then it skips this screen and goes to the book selection screen described further on down.

Image

In the screen above, each book collection is displayed.  The name of the book collection is displayed along with a count of "Downloaded # of #".  The total is actually the total of audio chapters for that book included as assets plus the number of audio chapters that are downloadable.  The number downloaded is actually the number included in the assets for the app plus the number that have been downloaded.  The "progress bar" underneath the totals is a visual representation of those numbers.  The arrow key takes you to the book selection screen.  (I'm using a couple of different apps for this, so the colors and fonts may vary, but it doesn't affect the functionality)

Image

The book collection screen shows each of the books that have "downloadable audio".  It is similar to the book collection screen with totals and progress bars.  The arrow key will allow the user to select specific chapters if they choose to do so.  But the downloads and deletions for an entire book can be done here.  The user can press box for the book or can press the "Select All" box to select or deselect all books.  Note that initially, when no items are checked that the top right of the navigation is empty.

Image

When the user presses "Select All" all book entries and the box for the "Select All" are checked.  If individual books are pressed after the "Select All" is pressed, that book is deselected and the "Select All " is unchecked.  Once one or more books have been checked, the delete and download buttons are displayed in the top right corner of the screen.  If the delete button is pressed, all downloaded audio files (but not those included as app assets) are deleted for the selected books and the totals are adjusted to reflect either 0 or the count of asset audio chapters that are loaded.   

Image

If the download button is pressed, all the downloadable chapters that have not been downloaded are downloaded, one book at a time, in order starting at the top.  The text under all of the selected boxes changes to "Waiting to download..." While the files for a book are downloading, the text under the button changes to "Downloading" and the count and progress bar are updated as each chapter completes downloading.  There is also a "Halt downloading" button that is displayed in the right corner while downloading, in place of the download and delete buttons.   

Image

 If the user presses the halt button, it does not immediately halt.  It displays a verification screen.  If the user answers yes to that, then the download stops.  Any files already downloaded are kept.

Image

If the user presses the arrow box next to the book, the chapter display screen is shown.  An entry is displayed for each screen containing a "downloadable audio" chapter file.  If the audio file is included in the assets, then it is shown as "Contained in app" with no box and the checkbox is checked.  Downloaded files are shown along with the file size.  There is a "Select All" button that behaves pretty much as it did at the last level, except that asset files cannot be selected or deselected.  A checkmark also appears for each chapter when it is completely downloaded.

Image

The only other difference is on the download screen.  The "Halt Download" button appears and functions the same way it did at the higher level.  However, while the chapter is downloading, the progress is updated as the download progresses, with the percentage complete reflected in the text and in the progress bar.

With the native apps, the files are placed in separate folders for each book collection and folder.  When an entry has been downloaded, an entry is made in a database table with the book collection id, book id, chapter number, the filename, the folder name, the date the file was downloaded and the file size.  It uses this when creating the screens.  When creating the chapter list screens, it validates that each file in the list is still present.

TheNonPirate:
Closes #1036 and #1112. This PR implements the download manager, from which the user is able to view which audio files have been downloaded, download any files they want, and delete files they do not want.

## Summary by CodeRabbit

* **New Features**
  * Added a Downloads page for browsing, downloading, deleting, and managing audio chapters.
  * Added download progress tracking, completion statuses, and cancellation controls.
  * Added a Downloads navigation item and new download-related icons.
  * Added confirmation prompts before canceling active downloads.
* **Improvements**
  * Audio downloads can now automatically start playback after successful completion.
  * Download failures are reported clearly, preventing playback until audio is available.
  * Audio alert modals can display localized titles.


src/lib/components/AudioDownloadModal.svelte
```diff
@@ -55,12 +58,22 @@ Audio Download Modal Dialog component.
         localStorage.setItem(STORAGE_CHOICE_KEY, handle ? 'enabled' : 'declined');
         await proceedWithDownload();
     }
-    export async function downloadAudio(url: string): ReturnType<typeof addAudioFile> {
+    export async function downloadAudio(
+        url: string,
+        item: { docSet: string; collection: string; book: string; chapter: string },
+        options?: {
+            noAutoplay?: boolean;
```
FyreByrd: Usually it would be better to name boolean variables in the positive, rather than in the negative.

src/lib/data/audio.ts
```diff
@@ -997,6 +1018,91 @@ export async function checkAudioAvailability(options?: { afterDownload?: () => v
...
+            },
+            audio
+        );
+        if (!foundAudioFile) {
+            let audioPath = '';
+            if (audioSource?.type === 'download') {
+                audioPath = pathJoin([audioSource.address, audio.filename]);
+            } else if (audioSource?.type === 'fcbh') {
+                if (!get(appOnline)) {
+                    modal.open(ModalType.AudioAlert, { messageKey: 'Audio_Download_Connect' }); //It looks like we'll actually need a slightly different audio alert (Should probably still go in the same modal) that uses Download_Check_Internet_Connection and also has a header with Notification_Channel_Name_Download
```
FyreByrd: Feel free to make this modification as part of this PR

TheNonPirate: That's the plan. Thanks for the reminder.

src/lib/data/audio.ts
```diff
@@ -997,6 +1022,106 @@ export async function checkAudioAvailability(options?: { afterDownload?: () => v
...
+                afterDownload: (success: boolean) => {
+                    downloadWorker.postMessage({
+                        type: 'FINISH_DOWNLOAD_AUDIO_ITEM',
+                        success
+                    });
+                }
+            }
+        );
+    }
+});
```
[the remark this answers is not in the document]
TheNonPirate: @FyreByrd Is this something we need to worry about, or because is SSR not an issue because we run everything clientside?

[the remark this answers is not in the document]
TheNonPirate: We don't do Server-Side Rendering.

FyreByrd (approved):
Looks good to me! Thanks for all the hard work!

PR #1119, opened 2026-08-13
Toggle font selection on verse on image

### LINKED ISSUE #1114: Verse on Image: do not show font chooser if there is just one
In the Verse on Image editor, if there is just one font, the Font tab can be hidden. There is no choice to be made.

Awest882:
* Only show font selector if there are more than 1 font options
* Closes #1114 

## Summary by CodeRabbit

* **Bug Fixes**
  * The font selection controls are now shown only when multiple font options are available, reducing unnecessary interface elements.
* **Maintenance**
  * Updated project test data formatting without changing its content.


test_data/projects/index.json
FyreByrd, line 1: Why was this file altered?

src/lib/components/VerseOnImage.svelte
```diff
@@ -118,7 +119,7 @@ The verse on image component.
 
     const optionIcons = [
         ImageIcon.Image,
-        ImageIcon.FontChoice,
+        ...($fontChoices!.length > 1 ? [ImageIcon.FontChoice] : []),
```
FyreByrd: Replace the source of `$fontChoices` with
```ts
const configFonts = [...new Set(scriptureConfig.fonts?.map((x) => x.family))];

export const fontChoices = derived(refs, ($refs) => {
    if (!$refs.initialized) {
        return configFonts;
    }
    const collection = scriptureConfig.bookCollections?.find((x) => x.id === $refs.collection)
    const bookFonts = collection?.books.find((x) => x.id === $refs.book)?.fonts;
    const currentFonts = bookFonts?.length ? bookFonts : collection?.fonts?.length ? collection.fonts : configFonts;
    return currentFonts;
});
```
Then remove the non-null assertion (`!`) from here and below.

src/lib/components/VerseOnImage.svelte
```diff
@@ -938,6 +939,7 @@ The verse on image component.
         "
     >
         <!-- NavBar of tab buttons to bring up the different editor panes -->
+        <!-- TODO HERE-->
```
FyreByrd: If there's still stuff to do, then this PR isn't ready for review... Please remove this unless there's more left.

FyreByrd (changes requested):

test_data/projects/index.json
Awest882: Found why it was changed, my auto-format removed a blank line at the end when I reverted back to it's original version

FyreByrd (approved):
Looks good to me. Thanks!

